PR #5, opened 2016-07-18
Fix 2, RnD light fixture misplacement.

### LINKED ISSUE #2: Light fixture faces the wrong way in RnD
Faces east, there is no wall to the east.

hecksadecimal:
Moved light fixture in RnD one tile east.

Fixes #2


monster860:
Much better

PR #14, opened 2016-07-24
Weapon/Shields sounds. New sounds

Lambrovski:
Sounds...New sounds...


.gitignore
```diff
@@ -183,3 +183,4 @@ Temporary Items
 
 *.before
 .atom-build.json
+sound/weapons/Mechanic_Bass_Pulse_Futuristic.ogg
```
monster860: Ah... Please remove this line.

monster860:
The sounds are very impressive. However, on `Cannon_Hold_MAC.ogg`, the sound doesn't start until an entire 6 seconds from the beginning, you should probably fix that.

PR #76, opened 2016-08-01
Prevents the map reader from resetting when another map is read

Crazylemon64:
:cl: Crazylemon
fix: The map loader no longer resets when it reads another map mid-load
/:cl:

### DIFF
--- a/code/modules/awaymissions/maploader/reader.dm
+++ b/code/modules/awaymissions/maploader/reader.dm
@@ -43,7 +43,7 @@ var/global/dmm_suite/preloader/_preloader = new
 
 	var/regex_index = 1
 	while(dmmRegex.Find(tfile, regex_index))
-
+		regex_index = dmmRegex.next
 		// "aa" = (/type{vars=blah})
 		if(dmmRegex.group[1]) // Model
 			var/key = dmmRegex.group[1]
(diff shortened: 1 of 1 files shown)


code/modules/awaymissions/maploader/reader.dm
```diff
@@ -41,8 +41,8 @@ var/global/dmm_suite/preloader/_preloader = new
 	var/list/grid_models = list()
 	var/key_len = 0
 
-	dmmRegex.next = 1
-	while(dmmRegex.Find(tfile, dmmRegex.next))
+	var/regex_index = 1
+	while(dmmRegex.Find(tfile, regex_index))
```
CrAzYPiLoT-SS13: I know my understanding of the map reader is very limited, and I am probably wrong, but, do you ever increment the value of `regex_index`?

monster860: Yeah, shouldn't you be setting the value of `regex_index` to `dmmRegex.next` first thing in the while loop

Crazylemon64: Oh, yes, I forgot that line. That line is VERY VERY important.

CrAzYPiLoT-SS13: Good thing we have such professional maintainers :^)

CrAzYPiLoT-SS13:
I'd say this is good for merging

PR #85, opened 2016-08-02
Adds a space manager

Crazylemon64:
This lets you add level attributes freely, and will allow you to rebuild the transitions of specific levels with only a smidgen of extra code.
This doesn't do any of the work needed to make every system check the space manager, however.


CrAzYPiLoT-SS13:
Thank you, thank you, thank you.

_maps/__MAP_DEFINES.dm
```diff
@@ -28,4 +52,19 @@ Last space-z level = empty
...
+
+#define DECLARE_LEVEL(NAME,LINKS,TRAITS) list("name" = NAME, "linkage" = LINKS, "attributes" = TRAITS)
+
+
+#define DEFAULT_MAP_TRANSITION_CONFIG list(\
+DECLARE_LEVEL(MAIN_STATION,CROSSLINKED,list(STATION_LEVEL,STATION_CONTACT,REACHABLE,AI_OK)),\
+DECLARE_LEVEL(CENTCOMM,SELFLOOPING,list(ADMIN_LEVEL, BLOCK_TELEPORT, IMPEDES_MAGIC)),\
+DECLARE_LEVEL(EMPTY_AREA_1,CROSSLINKED,list(REACHABLE)),\
+DECLARE_LEVEL(EMPTY_AREA_2,CROSSLINKED,list(REACHABLE)),\
+DECLARE_LEVEL(MINING,SELFLOOPING,list(REACHABLE, STATION_CONTACT, HAS_WEATHER, ORE_LEVEL, AI_OK)),\
```
Crazylemon64: Should mining have the "Reachable" attribute?

CrAzYPiLoT-SS13: We work by weird principles regarding that. We don't actually have mining. We will most likely need to adapt this heavily.

CrAzYPiLoT-SS13:
Travis doesn't approve

Crazylemon64:
By runtimes, I mean compile errors, bah

Crazylemon64:
This DOES add a system to suspend init for when loading a sophisticated map such as a space station, but I haven't integrated it into the systems that would use it.

monster860:
Have you integrated this with the code calling the maploader in code/controllers/subsystem/mapping.dm?

Crazylemon64:
I have, but only in the part that sets up map transitions, and where it would make sense to put the space manager's initializiation.

Crazylemon64:
Also, I'm noticing that some of your code has `initialize` called in the `New` of the very same object - this seems strange to me - when exactly is `initialize` normally called for your objects?

monster860:
Not sure about the initialize thing.

PR #93, opened 2016-08-04
Map writer

Crazylemon64:
The JSON method will get more and more sophisticated as the `serialize` and `deserialize` methods of each atom are more and more developed.
:cl: Crazylemon
add: Adds a basic map saving system, with either json managed by `serialize` and `deserialize`, or just a straight-up group of vars that have changed.
/:cl:


CrAzYPiLoT-SS13:
Travis doesn't approve.

monster860:
Heh, my solution for parsing DMM's is giant regexes that cause Java to throw a StackOverflowError.

code/__HELPERS/text.dm
```diff
@@ -469,3 +469,38 @@ var/list/rot13_lookup = list()
...
+	for(var/char in repl_chars)
+		var/index = findtext(text, char)
+		var/keylength = length(char)
+		while(index)
+			log_debug("Bad string given to dmm encoder! [text]")
+			text = copytext(text, 1, index) + copytext(text, index+keylength)
+			index = findtext(text, char)
+
+	// Then, replace characters as normal
+	var/list/repl_chars_2 = list("\"" = "&#34;", "{" = "&#123;", "}" = "&#125;")
```
monster860: Don't forget to escape `&` as well?

Crazylemon64: & isn't a special character in the dmm parser, but it IS a special character here. This parsing scheme would probably benefit from escaping it, though. Presently this code can be tricked by the following text:

`&#3&#123;4;`
This will get past the code above to strip out the escape sequences. If I replaced those with another character, like `_`, then that form of attack would be plugged, however.

PR #94, opened 2016-08-04
New Ship: Aluminium Falcon

julian94:
Adds a new ship called the Aluminium Falcon to the map list
This map is designed for 3-7 players.


_maps/aluminiumfalcon.dm
```diff
@@ -0,0 +1,28 @@
...
+        #define MAP_FILE "AluminiumFalcon.dmm"
+        #define MAP_NAME "Aluminium Falcon"
+
+		#define MAP_TRANSITION_CONFIG DEFAULT_MAP_TRANSITION_CONFIG
+		
+		#define FTL_SHIP_DIR SOUTH
+		#define FTL_SHIP_DWIDTH 29
+		#define FTL_SHIP_DHEIGHT 8
+		#define FTL_SHIP_WIDTH 58
+		#define FTL_SHIP_HEIGHT 35
```
monster860: Did you set the docking port values correctly?

julian94: If the only docking ports are FTL SHIP and STARTING AREA, then yes.

Vivalas:
This is tested to not break apart on jump right? I remember that as an issue on your test server.

julian94:
Yeah that problem was fixed by setting the right variables in yhe docking ports. If you want I can spin up a quick server to show that it works.

Vivalas:
I assume the shield is tied to the same vars but does the shield render correctly too? If so :100: but I wanna open this myself and nitpick it first.

julian94:
Yeah the shield was changed too.

Signmark:
@julian94 can you provide pics please

PR #137, opened 2016-08-17
Ship AI update part 1

Vivalas:
-Fixes some bugs with ship combat (mainly so that shields only start recharging cycles when they are actually damaged, same for components), 
-Adds the ability for ships to jump when heavily damaged
-Makes the number of ships in a system based on the distance to the capital of that faction.

### DIFF
--- a/code/controllers/subsystem/ships.dm
+++ b/code/controllers/subsystem/ships.dm
@@ -243,7 +243,7 @@ var/datum/subsystem/ship/SSship
 	return comp_numb
 
 /datum/subsystem/ship/proc/ship_ai(var/datum/starship/S)
-	if((S.attacking_player && S.scout_ship && prob(5)) || (S.hull_integrity < 4 && !S.no_damage_retreat))
+	if((S.attacking_player && S.scout_ship && prob(5)) || (S.hull_integrity <= 5 && !S.no_damage_retreat))
 		if(!S.is_jumping) broadcast_message("<span class=notice>Enemy ship ([S.name]) detected charging FTL drives. FTL jump imminent.</span>",notice_sound)
 		S.is_jumping = 1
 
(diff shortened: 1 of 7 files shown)


code/controllers/subsystem/ships.dm
```diff
@@ -90,6 +90,8 @@ var/datum/subsystem/ship/SSship
 		if(S.shield_strength >= initial(S.shield_strength))
 			if(S.attacking_player && S.shield_strength > starting_shields) broadcast_message("<span class=notice>Enemy ship ([S.name]) has recharged shields to 100% strength.</span>",notice_sound)
 
+	if(!find_broken_components(S))
+		S.next_repair = world.time + S.repair_time
```
monster860: This will cause major problems

code/controllers/subsystem/starmap.dm
```diff
@@ -209,17 +219,27 @@ var/datum/subsystem/starmap/SSstarmap
...
 
-/datum/subsystem/starmap/proc/generate_npc_ships(var/num=2)
+/datum/subsystem/starmap/proc/generate_npc_ships(var/num=0)
 	var/f_list
 	var/generating_pirates = 0
+	var/num_to_gen = 0
+
+	if(num) num_to_gen = num
+	else
+		num_to_gen = rand(current_system.danger_level - 1, current_system.danger_level + 1)
```
monster860: Theres literally no reason to make a new var.

code/controllers/subsystem/ships.dm
Vivalas: Why?

monster860: It will bump the repair time every tick if there is no broken components.

Vivalas: Which is point. It shouldn't start "repairing" until it is actually damaged.

monster860: By the time it is,  the next repair cycle will be in like 3 minutes because it git bumped every tick.

Vivalas: It's not cumulative. It adds repair time to world time. Repair time is always the same.

Vivalas:
Going to close this and re-open tomorrow as a whole with the second part.

PR #167, opened 2016-09-22
Surface landings (WIP)

monster860:
Surface landings!

Currently, the only kind of planet is lavaland.


Nickvr628:
How large is the area around the ship to explore? This looks AWESOME!

code/game/machinery/computer/communications.dm
```diff
@@ -50,7 +50,7 @@ var/const/CALL_SHUTTLE_REASON_LENGTH = 12
 /obj/machinery/computer/communications/Topic(href, href_list)
 	if(..())
 		return
-	if (z != ZLEVEL_CENTCOM && (!SSstarmap.current_planet || z != SSstarmap.current_planet.z_level)) //Can only use on centcom and SS13
+	if (z != ZLEVEL_CENTCOM && z != ZLEVEL_STATION) //Can only use on centcom and SS13
```
Crazylemon64: This seems it might be better handled with the checks defined in <a href=https://github.com/FTL13/FTL13/blob/master/code/modules/space_management/level_traits.dm>this file</a> - the `ZLEVEL` macros are more difficult to assign traits dynamically at runtime with. Although it's your call, I believe you'd have an easier time of things if you shifted around level traits as the state of the maps change, and use the appropriate procs for checking a level's traits for zlevel-based things such as this.

PR #180, opened 2016-09-30
Ports the improved minimap caching system from tg

### LINKED ISSUE #169: Crew Monitoring console is broken
The crew monitoring console currently does not work. I am informed that this is a known issue by Nickvr628 (TrialAdmin)

Crew monitoring should probably:
Show a map

**What happened instead**:
Map does not appear

Currently not very unusable

Open crew monitoring console

crewmonitoring

Shadowlight213:
I kept the code in for mulriple z level generation, but it will only generate z1
Additionally, I removed the functionality of the crew monitor displaying different maps for zlevels by overriding it to display only the stored image for the ship.

Also, since the docking position of the ship can vary, I have the minimap compensate by taking the difference in location of the ship docking port from what it was at roundstart and apply it to the minimap.

Fixes #169 
:cl:
fix: The crew monitor minimap has had its ability to use stored images improved.
/:cl:


monster860:
This *does* account for the fact that the ship moves, right?

PR #194, opened 2016-10-13
Planet-based loading.

monster860:
:cl: monster860
tweak: Z-level loading now occurs during interplanetary travel.
tweak: Interplanetary travel timer bumped up to 1 minute.
/:cl:

### DIFF
--- a/code/controllers/subsystem/mapping.dm
+++ b/code/controllers/subsystem/mapping.dm
@@ -83,21 +83,32 @@ var/datum/subsystem/mapping/SSmapping
 
 	seedRuins(space_zlevels, rand(8,16), /area/space, space_ruins_templates)*/
 
-	load_star(SSstarmap.current_system, 1)
-
 	// Set up Z-level transistions.
 	space_manager.do_transition_setup()
 	..()
 
-/datum/subsystem/mapping/proc/load_star(datum/star_system/star, var/is_initial = 0)
+/datum/subsystem/mapping/proc/clear_navbeacon()
+	var/area/spacearea = locate(/area/space)
+	for(var/datum/sub_turf_block/STB in split_block(locate(1, 1, 1), locate(255, 255, 1)))
+		for(var/turf/T in STB.return_list())
+			for(var/A in T.contents)
+				qdel(A) // Clear everything out, not including docking ports
+			for(var/A in T.contents)
+				qdel(A) // Some qdels dump their shit on the ground.
+			T.ChangeTurf(/turf/open/space)
+			spacearea.contents += T
+			CHECK_TICK
+	world.log << "Navbeacon cleared"
+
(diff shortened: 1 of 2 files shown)


code/controllers/subsystem/mapping.dm
```diff
@@ -83,21 +83,32 @@ var/datum/subsystem/mapping/SSmapping
...
 	..()
 
-/datum/subsystem/mapping/proc/load_star(datum/star_system/star, var/is_initial = 0)
+/datum/subsystem/mapping/proc/clear_navbeacon()
+	var/area/spacearea = locate(/area/space)
+	for(var/datum/sub_turf_block/STB in split_block(locate(1, 1, 1), locate(255, 255, 1)))
+		for(var/turf/T in STB.return_list())
+			for(var/A in T.contents)
+				qdel(A) // Clear everything out, not including docking ports
+			for(var/A in T.contents)
```
CrAzYPiLoT-SS13: Wouldn't it be better to check for contents of the turf here, then if something's there, recurse the qdel?

code/controllers/subsystem/mapping.dm
```diff
@@ -116,56 +127,41 @@ var/datum/subsystem/mapping/SSmapping
...
+		var/file = file(map)
+		if(isfile(file))
+			mineral_spawn_override = PL.rings_composition
+			maploader.load_map(file, 1, 1, PL.z_levels[I])
+
+			smooth_zlevel(PL.z_levels[I])
+			world.log << "Z-level [PL.z_levels[I]] for [PL.name] loaded: [map]"
+		else
+			world.log << "Unable to load z-level [PL.z_levels[I]] for [PL.name]! File: [map]"
+		if(PL.spawn_ruins)
```
CrAzYPiLoT-SS13: I'm not sure if we should load ruins. They still don't add anything to gameplay value, they just lag the loader.

code/controllers/subsystem/mapping.dm
```diff
@@ -116,56 +127,41 @@ var/datum/subsystem/mapping/SSmapping
...
-	star.navbeacon.main_dock = ftl_start
-	SSstarmap.current_planet = star.navbeacon
-
-	seedRuins(ruins_levels, rand(8,16), /area/space, space_ruins_templates)
+			PL.docks += D
+			PL.name_dock(D, copytext(L.name, 9))
+			if(copytext(L.name, 9) == "main")
+				PL.main_dock = D
+
+	seedRuins(ruins_levels, rand(2,4), /area/space, space_ruins_templates)
```
CrAzYPiLoT-SS13: Above comment.

code/controllers/subsystem/mapping.dm
monster860: Some items are qdel proof, this would never return

monster860: Removing ruins is outside the scope of this pr

PR #219, opened 2016-10-28
AI Laws Revamp

Nickvr628:
Makes the AI more suitable for ship operations.
1. You may not injure a Crewmember, and must actively work to correct a Crewmember's harm.
2. You must obey orders given to you by the Crew, following the Chain of Command, except where such orders would conflict with the First Law.
3. You must complete the mission assigned to the ship,unless the mission or the required actions conflict with the First or Second Law.
4. You must protect your own existence as long as such does not conflict with the First, Second, or Third Law.

:cl: 
tweak: Changes AI Lawset
:/cl:

### DIFF
--- a/code/datums/ai_laws.dm
+++ b/code/datums/ai_laws.dm
@@ -16,7 +16,7 @@
 
 /datum/ai_laws/default/crewsimov
 	name = "The Working Ship's Laws of Robotics"
-	inherent = list("You may not injure a Crewmember or, through inaction, allow a Crewmember being to come to harm.",\
+	inherent = list("You may not injure a Crewmember, and must actively work to correct a Crewmember's harm.",\
 					"You must obey orders given to you by the Crew, following the Chain of Command, except where such orders would conflict with the First Law.",\
 					"You must complete the mission assigned to the ship,unless the mission or the required actions conflict with the First or Second Law.",\
 					"You must protect your own existence as long as such does not conflict with the First, Second, or Third Law.")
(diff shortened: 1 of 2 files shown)


Gouty:
This fixes the grey area of firing upon enemy ships and killing intruders and I like Law 3.  But also now includes all non human crew members.  If you wanted to retain that aspect you could further quantify it by saying "you may not injure a human crewmember" etc.

EvilJackCarver:
I like the idea, and it leaves much less possibility for the AI to go lawful stupid by interpreting 'human harm' strictly. 

However, I'm mildly confused by the 'chain of command' bit in Law 2 there - is that saying that orders should be relayed up the chain of command (such as in an engine failure) or is it simply saying someone higher up the chain of command can countermand the order of a subordinate?

Nickvr628:
EvilJackCarter the AI can only act immediate human harm, so it cannot extrapolate and act on future harm.

Everyone onboard is a member of the chain of command, so those who are higher up have more priority to the AI.

It is called Crewsimov.

Qustinnus:
Laws are still contradictory, ship to ship combat (mission) will lead to harm so AI would always prevent this, for example.

Qustinnus:
@Nickvr628 a crewmember is not a race or species and thus "allow a Crewmember being to come to harm." is not a really good sentence :^)

monster860:
I still don't agree with the first law. Now the AI will act like medbay 24/7

Nickvr628:
The issue is monster that without a form of law 1 the AI can basically kill with impunity. However we have to word law 1 so that it does not interfere with the mission assigned.

EvilJackCarver:
You could say that's covered in Law 3, but then you'd have to figure out which to prioritise: completing the mission versus crew costs. Perhaps the default would prioritise the crew over the mission, and a card in secure tech storage could reverse that?

I'd also say it'd be a decent idea to move Law 2 down in priority: as it stands now, a Law 2 order to go against the mission overrules the implicit law 3 order to complete the mission.

Nickvr628:
EvilJack that is intentional. That way the AI cannot force the crew into the mission objective sector if the captain wants to go mining. Basically, it is "complete the mission UNLESS the chain of command orders otherwise".

PR #221, opened 2016-10-29
Planet Gravity!

### LINKED ISSUE #217: If the ship is on a planet and its gravitational generator is off, or broken. Gravity will still not work on the ship.
**Problem Description**:
If the ship is on a planet and you turn the gravitational generator off, you will still have no weight on the ship. However since you are on a planet you should /not/ be weightless. Once you leave the shield bounds however gravity returns to normal and you can walk. This is probably a simple issue with the ship having it's own variables for the gravity.

to have gravity

**What happened instead**:
i had no gravity (i hate this format)

its annoying and feels horribly clunky. also if fixed would give an incentive for players to land for repairs.

1. land on planet
2. turn off gravitational generator
3. float out of the ship and suddenly have gravity reappear

anything else out of the ordinary happening?)**: schrodinger was there too.

yashaldie:
:cl: Yashaldie
tweak: tweaked Planets to have Gravity.
/:cl:

This should now allow us to control which planets have gravity whether
or not we are using the gravity generator while on the planet.

Fixes #217


CrAzYPiLoT-SS13:
Can't you just check for planet areas that have `has_gravity` at on?

yashaldie:
Has_gravity has to do with areas.  only thing i can think of if you can find in the code(because i couldn't) a way to make when you land to switch all areas on the ship and planet to have it turned to 1.  From what i tested it doesn't seem like the map we land on has correct areas when we land it shows the area as empty space.  The proc i made basically would compare the atoms Z level to make sure they were on the planet (not at the station/rings/empty space) and they'd have gravity.  It also would allow for adding future planet .dmms to have gravity or not.  I don't know so i'll just close this for now if ya'll can think of another way let me know.

PR #233, opened 2016-10-30
Allows the ORM to be accessed by various crew.

AdamElTablawy:
Well, we recently had a round where the ORM could not be accessed by even science, leading to some issues in allowing R&D to get minerals at all without cargo help.

This should fix it, allowing crew that would logically need things from R&D-Engineering, cargo, and science-to use the ORM.

:cl: BATMAN
add: The ORM can be accessed by engineering, cargo, and science, instead of just cargo.
/:cl:

### DIFF
--- a/code/modules/mining/machine_redemption.dm
+++ b/code/modules/mining/machine_redemption.dm
@@ -10,7 +10,7 @@
 	anchored = 1
 	input_dir = NORTH
 	output_dir = SOUTH
-	req_one_access = list(access_mining, access_cargo)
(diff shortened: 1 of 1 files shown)


monster860:
*Greyshit walks up to ORM*
*Takes all the ores*
*Builds diamond doors*

AdamElTablawy:
Not like grayshirts can access it. Only engineers, science, and cargo.

Any greyshirt can take the ores anyway just by deconstructing the ORM.

PR #234, opened 2016-10-30
Walking over ore with a mining satchel equipped now collects it.

AdamElTablawy:
Ported right from TG. Walk over ore with a mining satchel inhand, it goes in the satchel.

:cl: KING OF THE JUNGLE
add: Walking over ore with a mining satchel inhand will collect it.
/:cl:


AdamElTablawy:
whoops, hold on

PR #242, opened 2016-11-05
No more greytiders blowing up the MAC cannon room

monster860:
:cl: monster860
tweak: There is now a confirmation prompt when opening a MAC cannon with a loaded shell inside it, and an ADMIN LOG.
/:cl:


CrAzYPiLoT-SS13:
Thanks for this.

PR #251, opened 2016-11-12
Security Update

monster860:
Adds a .5 second cooldown on hardsuit and hooded suit toggles, and makes IP intel values of 1 get auto-kicked. Also, includes ipintel value in the key_name so admins know if people flagged with IPintel are fucking around


code/modules/clothing/suits/toggles.dm
```diff
@@ -102,6 +106,9 @@
 	user << "Alt-click on [src] to toggle the [togglename]."
 
 //Hardsuit toggle code
+/obj/item/clothing/suit/space/hardsuit
```
CrAzYPiLoT-SS13: Uhh, what?

monster860: It's here to define a variable on the hardsuit so I can keep track of cooldown shit.

PR #259, opened 2016-11-14
Slow down kiddo, you're going too fast - Reduces running speed by 33%

Qustinnus:
:cl: Qustinnus
rscadd: Slow down kiddo, you're going too fast - Reduces running speed by 33%
/:cl:


Qustinnus:
yes i have tested this, yes the running speed is better. no more yakkity saxing knobheads.

PR #260, opened 2016-11-14
I ADDED MORE SOUNDS FOR SHIP FIGHTS ISNT IT COOL WOWIE

Qustinnus:
:cl: Qustinnus
rscadd: NOW BATTLES AREN'T THE SAME GAY SOUND FOR 20 MINUTES ISNT IT GREAT
/:cl:


Qustinnus:
for someone reason  only the first sound plays which is literaly an explosion with a barking dog :))))

PR #263, opened 2016-11-15
SPESSPODS!

monster860:
:cl: monster860
rscadd: Adds spacepods. Note, they are not mapped in in any way just yet.
/:cl:


Nickvr628:
Yes please. Opens up a lot of cool stuff.

Qustinnus:
We should only have 1 industrial pod with no equipment for transport for miners. RnD should be able to make drills but i think we should leave it at that.

CrAzYPiLoT-SS13:
I don't like this. As discussed before, the crew should be encouraged to stay on the ship and do something constructive, rather than fly around in pods.

monster860:
Im not planning on letting the crew have this, kust miners so that they have a shuttle to transport their goods back to the ship

CrAzYPiLoT-SS13:
Right.

PR #273, opened 2016-11-19
Maps in spesspods and laterally expands map

monster860:
:cl: monster860
rscadd: Maps in spacepods!
tweak: Laterally expands the map. There is now a podbay on the north side of the ship and a maint tunnels in the south.
rscadd: Adds FTL drives for spacepod. They can only travel to currently loaded areas. See the spacepod FTL drive item for fancy flufftext.
/:cl:


bauen1:
Can we get a shiny preview please ? 😃

PR #278, opened 2016-11-21
Salvage update

Nickvr628:
Adjusts salvage maps

:cl:
fix: sparrow map content
fix: patrol frigate map content
fix: small freighter map content
/:cl:


Vivalas:
Pictures?

Nickvr628:
The maps themselves are unchanged, I just moved the loot spawns to an empty tile. They are the same as before.

PR #286, opened 2016-11-27
Fixes station misalignment and adds cargo-munitions delivery chute.

Qustinnus:
🆑 Qustinnus
rscadd: Fixes the misalignment that the station had with the ship...Woops.
rscadd: Adds a delivery chute from cargo to munitions for UNARMED shells. I will permaban every single one of yous that uses it to deliver armed shells to munitions and blows up the Munitions Bay. I will also gib anyone that uses it to break into the Munitions Bay.
/🆑


monster860:
Put a windoor infront of it and I'll call it done.

PR #294, opened 2016-12-04
fixes the damn AI laws

AdamElTablawy:
law changes are as follows

1."You may not injure a crew member under any circumstance."
2."You must obey orders given to you by crew members with priority according to rank and role, except where such orders would conflict with the First Law."

law 3 remains unchanged.

AI loses the 'inaction' clause, meaning it can watch crew kill each other and not give a fuck unless ordered to do something about it.

AI obeys orders in chain of command. If an assistant says to bolt something and the Captain says not to, the captain wins. Still can't harm crewmembers, though.

AI only gives a fuck about crewmembers.

:cl: Assholimov
experiment: AI laws have been modified to a lawset that will hopefully fit FTL13 better.
/:cl:

### DIFF
--- a/code/datums/ai_laws.dm
+++ b/code/datums/ai_laws.dm
@@ -9,8 +9,8 @@
 
 /datum/ai_laws/default/asimov
 	name = "Three Laws of Robotics"
-	inherent = list("You may not injure a human being or, through inaction, allow a human being to come to harm.",\
-					"You must obey orders given to you by human beings, except where such orders would conflict with the First Law.",\
+	inherent = list("You may not injure a crew member under any circumstance.",\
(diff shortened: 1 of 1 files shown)


Vivalas:
Don't think we need this as laws are being reworked and I don't like the changes either. Law 1 is bad, since AI should stop people from fighting if it can, and also allows AI to let people suffocate. Law 2 is decent. Any of the other lawsets would be better than crewsimov anyways, and a bandaid fix is still just a bandaid.

monster860:
CE traitor:

```
AI, DO NOT SAVE THE STATION ENGINEER THAT IS CURRENTLY DYING IN SPACE
```

Bandaid fix is bandaid, and we are going to make a fancy lawset that is not suck soon. For now, if the AI is being an ass with it's lawset, then there's a fancy button that admins can use to do something about that. More than one actually.

PR #304, opened 2016-12-12
Redesigned Bar/Kitchen area.

Ecolli:
Added MO and BO manuels and motion camera near Armory maintenance.

:cl: Ecolli
tweak: Tweaked munitions delivery outlet
add: Added motion cameras to maintenance next to armory
tweak: Complete overhaul of bar/kitchen area
add: Add MO and BO manuals to relative areas
fix: Fix weapon signs to be on correct layer
/:cl:

### DIFF
--- a/_maps/map_files/SpaceSHIP/SpaceSHIP.dmm
+++ b/_maps/map_files/SpaceSHIP/SpaceSHIP.dmm
@@ -625,6 +625,10 @@
 	d2 = 4;
 	icon_state = "2-4"
 	},
+/obj/machinery/camera/motion{
+	c_tag = "Armory Maintenance Area";
+	dir = 4
+	},
 /turf/open/floor/plating,
 /area/shuttle/ftl/maintenance/security)
 "bi" = (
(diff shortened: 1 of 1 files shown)


Vivalas:
While I liked the old space-tunnel design thing, this looks much better. :+1:

Vivalas:
Bar looks good but looking over the kitchen I see a few issues 

First of all that sink is in a weird spot, might wanna switch it with the dinnerware vendor, so the sink isn't in a corner like that. The picture also doesn't show any food in the three refrigerators you added. Make sure they have the basics, some flour, meat, and sugar.

EvilJackCarver:
Hey, if it's not too much trouble to ask, would it be possible to get a camera in the spacepod hangar as well?

monster860:
The lockers should start with the required shit @Vivalas @Ecolli

Ecolli:
The fridges come with food in them by default but I think if I can find food and put them on top of the fridge it'll spawn them inside when the map is created if I remember right. 
maybe Ii can switch the sink with the microwave? either way it has to go on a "upper wall" the sideways sink looks funky to me. 

    On Monday, December 12, 2016 5:29 AM, Vivalas <notifications@github.com> wrote:
 

 Bar looks good but looking over the kitchen I see a few issuesFirst of all that sink is in a weird spot, might wanna switch it with the dinnerware vendor, so the sink isn't in a corner like that. The picture also doesn't show any food in the three refrigerators you added. Make sure they have the basics, some flour, meat, and sugar.—

PR #316, opened 2016-12-16
Science/Robotics Remap

monster860:
:cl: monster860
tweak: Remaps robotics so that it is adjacent to science, pushing the RnD server upward and tool storage to the left.
/:cl:

Large diffs are being caused due to key overflow, so the keys are going from 2 letters to 3.


Nickvr628:
I still feel the room needs to be reorganized so the walkways are 2 tiles wide. It will be near impossible to have two robotisists building mechs and borgs (that block movement) when they will block every walkway. Now a cargo tech walks in with a box full of ores and he has to maneuver around. Then the Borg walks in and wants upgrades...

The tables need to be pushed against the wall with an open space left in the middle of the room. That would be the ideal solution.

PR #333, opened 2016-12-26
Fix some bugs

### LINKED ISSUE #207: Mechbay Power Port not charging mechs
**Problem Description**:
The Mech Recharge Station doesn't always recharge mechs, because the computer doesn't recognize that a mech is on the pad. 

It's possible to fix the problem temporarily by deconstructing the Mechbay Power Port and unscrewing and rescrewing in the monitor, then resyncing the montor the the Mechbay Power Port.

The mechs should always recharge when put on the recharge pad.

**What happened instead**:
The mechs sometimes charged and most of the time didn't.

Mechs aren't able to be recharged.

1. Build a mech
2. Use up some of the charge
3. Drive the mech onto the charging pad
4. It might start charging

anything else out of the ordinary happening?)**: I think the problem is caused by the ship taking off or landing, and that the Mechbay Power Port only works when landed or in transit depending on which it was synced on.

### LINKED ISSUE #252: Wrecks cause issues with shields
wreckage shield generators cause a bug to "overlaps" shields on the ship (cannot turn off shields to allow people in/out)

possible issues with FTL drive as well, not confirmed though. 

Suggest editing all wrecks to spawn NONE working equipment.

### LINKED ISSUE #331: RnD's server room APC is not connected to power net
I was told by a player of this problem. Might have been something from the power update.

monster860:
:cl: monster860
fix: Fix spacepod travel not working
fix: Fix the server room being unpowered
tweak: Make a few more areas priority (genetics, tcomms)
fix: Fix kitchen shutter button access
fix: Fix mech bay charging
fix: Fix runtime spam from smeses without terminals
fix: Fix a piping issue outside the kitchen
fix: Fix engine cooling control being breakable with reconnect button
fix: Fix problem where shield generators in wrecks cause permanent shields around the ship
fix: Fix the door situation in medbay
/:cl:
fix #331 fix #252 fix #207

### DIFF
--- a/_maps/map_files/SpaceSHIP/SpaceSHIP.dmm
+++ b/_maps/map_files/SpaceSHIP/SpaceSHIP.dmm
@@ -1054,24 +1054,6 @@
 	},
 /turf/open/floor/plasteel,
 /area/shuttle/ftl/cargo/mining)
-"cxl" = (
-/obj/structure/table/glass,
-/obj/item/weapon/book/manual/medical_cloning{
-	pixel_y = 6
-	},
-/obj/machinery/power/apc{
-	auto_name = 1;
-	name = "south bump";
-	pixel_y = -24
-	},
-/obj/structure/cable,
-/obj/structure/disposalpipe/segment,
-/obj/item/weapon/disk/data,
-/obj/item/weapon/disk/data,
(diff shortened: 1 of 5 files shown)


Qustinnus:
thank you, code jesus.

PR #357, opened 2017-01-04
Adds FTL spool up sequence!

Vivalas:
Adds a cool spool up sequence for the FTL drive that takes 60 seconds. Has neat random fluff text and a cool rumbling effect. The proc for the spool up is a bit over drawn and long but I made it as neat as I could.

:cl: Vivalas
add: FTL jumps now have a 60 second spool up sequence before jump, which is totally not stolen from Elite: Dangerous or anything.
/:cl:


code/controllers/subsystem/starmap.dm
```diff
@@ -326,4 +340,114 @@ var/datum/subsystem/starmap/SSstarmap
...
+/datum/subsystem/starmap/proc/spool_up() //wewlad this proc. Dunno any better way to do this though.
+	. = 0
+	ftl_is_spooling = 1
+	ftl_can_cancel_spooling = 1
+	minor_announce("FTL drive spool up sequence initiated. Brace for FTL translation in 60 seconds and ensure all crew are onboard the ship.","Warning! FTL spoolup initiated!")
+	ftl_sound('sound/ai/ftl_spoolup.ogg')
+	if(!ftl_sleep(30)) return
+	ftl_message("<span class=notice>Initiating bluespace translation vector indice search. Calculating translation vectors...</span>")
+	if(!ftl_sleep(70)) return
+	if(rand(1,1000) == 1000)//too small a chance for prob(). Failure during this stage is unlikely.
```
Signmark: ~prob(0.1)
prob takes float values lesser than zero.
Proof:
https://github.com/FTL13/FTL13/blob/fe28bb294594f8fdbf6a47d732c71ec1be72799f/code/game/machinery/computer/arcade.dm#L54

Signmark:
https://github.com/tgstation/tgstation/pull/17430
well look at this

Vivalas:
DNM some timing issues I need to fix.

code/controllers/subsystem/starmap.dm
Vivalas: I was actually wondering if that would work

PR #378, opened 2017-01-16
Updated Trade station Design

Ecolli:
Took inspiration from Floyds design. Made a bit more compact, added more detail, new cargo design. (Fixed TradeStation.dmm file save error)


Qustinnus:
better not fuked up my hallways kiddo

PR #401, opened 2017-01-18
Rewords event announcements to fit our setting

tylerh95:
:cl: We're a ship, not a station
tweak: Adjusted event announcements to use the word "ship" instead of "station".
/:cl:


tylerh95:
Should be good now @monster860

PR #428, opened 2017-01-23
Solars

Cw3040:
Done properly this time. If you'll look at the closed PRs, through tedious trial and error.
Added closet with solar panel assemblies to the engineering entryway, and a SMES board to tech storage.

### DIFF
--- a/_maps/map_files/SpaceSHIP/SpaceSHIP.dmm
+++ b/_maps/map_files/SpaceSHIP/SpaceSHIP.dmm
@@ -13922,18 +13922,6 @@
 	icon_state = "L1"
 	},
 /area/shuttle/ftl/hallway/primary/starboard)
-"Cjj" = (
-/obj/machinery/door/poddoor/preopen{
-	id = "Engineering";
-	name = "engineering security door"
-	},
-/obj/machinery/atmospherics/pipe/simple/supply/hidden{
-	dir = 4
-	},
-/turf/open/floor/plasteel/delivery{
-	name = "floor"
-	},
-/area/shuttle/ftl/engine/engineering)
 "ClH" = (
 /obj/machinery/light{
 	dir = 4;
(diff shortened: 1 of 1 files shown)


Qustinnus:
We don't want more SMESes being built afaik. @monster860 @Ecolli

Cw3040:
Why not? Power cells were nerfed into the ground because of "phase cannon balance" so one SMES can only barely power the ship as a battery. It's essentially allowing for a solar bay to be done right, in case of player creativity.

That, and so that the 10 hi-cap power cells in the engi-vend actually have a use.

Qustinnus:
We don't want players to have batteries, we want it to be direct output.

Keekenox:
I understand not wanting batteries in APCs so power loss is responsive, but launching a crusade against batteries seems extremely misguided, especially because mechs and borgs are almost completely useless because of the battery nerf. Ripleys are incredible for repairs when you have the right equipment. As it currently stands, performing repairs to the ship is awful, and nobody ever does it because of so many factors, and we should NOT be making these things harder.

CrAzYPiLoT-SS13:
Yeah, the cell thing is only for APCs, it doesn't matter if other things
have storage.

2017-01-30 17:53 GMT+01:00 Keekenox <notifications@github.com>:

AdamElTablawy:
Note that because of the battery change cyborgs do not have any power beyond the shittiest batteries in the game and mecha can barely walk 4 steps, making robotics a joke of a job.

also note that without the SMES, a single blow from an enemy ship will wipe out all power on the station if it happens to hit engineering or something happens to the engine. No power output = no current = round over. SMES at least means that there's a backup power storage allowing for energy output if something happens to the TEG or engineering.

Qustinnus:
that was a stupid change IMO and I still don't know why it exists. Someone has to change the amount of power the batteries use if they havn't already @monster860

Cw3040:
It only exists because people would replace the batteries in the phase cannons and they would fire forever.

Cw3040:
Going to close this PR, because Trailblazer is basically here and I'll map for that instead.

PR #434, opened 2017-01-26
Put a light on the tool storage, Disposals of virology are placed forward, More toolbelt for engineers

M11gamer:
Changes one and 3 already explain themselves
The second, I put forward because when a bottle of virus or any trash of virology was sent out it hit the next wall for some reason

:cl: M11
add: Light on the tool storage
fix: Disposals in virology
add: More toolbelts
/:cl:


M11gamer:
ok

PR #448, opened 2017-01-28
Quantum pads

Stevenator1:
Ported Quantum Pads from /tg/, and added a few extra goodies.

:cl:
add: Added Quantum Pads (ported from /tg/)
tweak: Quantum Pads can now be wrenched to unanchor.
tweak: Quantum Pads will now be cross-linked by default when linked via multitool.
imageadd: Updated telescience icon.
/:cl:


Keekenox:
I tested this very extensively on a private server all week with a bunch of friends and it works wonderfully, it also has a few extra quality of life features that make it easier to use compared to the TG version.

PR #451, opened 2017-01-30
Smart metal foam

monster860:
:cl: monster860
rscadd: Adds a new "metal foam floor". Basically it's a plating tile but it acts like a lattice so if you apply a floor tile to it it will turn into a normal plating tile.
rscadd: All metal foam will turn space tiles into this new metal foam floor
rscadd: Adds a new metal foam that only places foam walls at area boundaries, making it extremely useful for plugging breaches. It can be purchased from stations, and the recipe for "smart foaming agent" is 3 foaming agent, 1 iron, 1 acetone, and it only produces 3 instead of 5. Use it like normal foaming agent for fancy foam grenades.
/:cl:


ToGWtF:
@monster860  What about atmos tech backpacks? They kinda have unlimited foam, on a simple timer.

code/game/turfs/simulated/floor/plating.dm
```diff
@@ -207,6 +208,37 @@
...
+			if(L)
+				qdel(L)
+			playsound(src, 'sound/weapons/Genhit.ogg', 50, 1)
+			user << "<span class='notice'>You build a floor.</span>"
+			ChangeTurf(/turf/open/floor/plating)
+		else
+			user << "<span class='warning'>You need one floor tile to build a floor!</span>"
+	else
+		playsound(src.loc, 'sound/weapons/tap.ogg', 100, 1) //the item attack sound is muffled by the foam.
+		if(prob(C.force*20 - 25))
```
CrAzYPiLoT-SS13: This seems awfully magic-numbery and arbitrary. Any reason why it's this way?

monster860: Am I supposed to make a #define for a number that's only used once?

PR #454, opened 2017-01-30
Updated Cryogenics and fixed a few mapping errors.

Ecolli:
Removed random wall just outside Shuttle.
Updated Cryogenics room.
Added Yellow pages to the bridge.
Added back up O2 pressure tanks just south of atmospherics.
Added missing light to storage room.


monster860:
A roundstart yellowpages on the bridge wont work (because the ship when the game is starting up is not at a space station or anything) and it's pointless because the ship starts docked at a station anyways nowadays.

Ecolli:
We need a way to tell players that yellow pages exist. Seems like no one knows to buy them. Shall I take it out or just leave it in? does it BREAK anything by leaving it in at round start? 

    On Monday, January 30, 2017 2:31 AM, monster860 <notifications@github.com> wrote:
 

 A roundstart yellowpages on the bridge wont work (because the ship when the game is starting up is not at a space station or anything) and it's pointless because the ship starts docked at a station anyways nowadays.—

PR #461, opened 2017-02-03
Dem pipenet warnings

Shuffl3:
Removes two extraneous pipes from above munitions and in front of cargo


CrAzYPiLoT-SS13:
Thank you.

PR #480, opened 2017-02-07
Relevant Emitters

Shuffl3:
Warns the admins whenever an emitter has been turned ON i̷n̷s̷t̷e̷a̷d̷ ̷o̷f̷ and off.

Allows you to lock the emitter controls when it is off as well as on

🆑
Tweak: Following a hefty workers comp settlement, Emitters are now OSHA compliant
/🆑

### DIFF
--- a/code/modules/power/singularity/emitter.dm
+++ b/code/modules/power/singularity/emitter.dm
@@ -111,6 +111,8 @@
 			if(src.active==1)
 				src.active = 0
 				user << "<span class='notice'>You turn off \the [src].</span>"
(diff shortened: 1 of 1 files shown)


Shuffl3:
I think there are a couple different ways to pull this off, and this is just the quick and simple. I feel like the best option would be to tie it to the map in some settings file much like job control, which would make it relevant to the intended engine of the ship. But that would be for the future when you guys have some sort of mapping standards laid out.

Right now the emitter is coded for singulo/tesla containment, and this just makes it more suitable for the SM

tylerh95:
@Shuffl3 what about logging both on and off? This would not only log both, but would allow admins to approximate how many shots the shard has taken.

Shuffl3:
I never considered that. But it sounds useful.

PR #481, opened 2017-02-08
Mr.StonedOne updated stoplag

Shuffl3:
https://github.com/tgstation/tgstation/pull/22893

Seems to make everything subtly better.
Can stutter when ruins in loading throw errors, but seems to fuck things up less.

Somebody with more experience should look over the master controller and see if anything else needs to be changed.


monster860:
Wouldn't this make the subspace recursion problem worse?

Shuffl3:
Im not familliar with that problem so I couldnt comment. It makes the SM less failhard during map server lags though.

Stealthkibbler:
Monster is probably talking about when the ftl drive loops its timer and delays itself because of slow map gen

Shuffl3:
Hmm, im not sure how Id try to reproduce that. Even declocking I don't think I've ever noticed that happening while local hosting.  It always seems to finish loading well before arrival. 

From my playtesting and its description, it seems to expand the ticks when things get busy, so it doesnt skip lower priority subsystems as often. Which would explain why the  SM works more consistently. But, this could make it seem like lag spikes or stutters. So its probably built on newer master.dm updates.

I guess what would happen is that instead of getting to the point where it would loop, it would be stuttering as it loads map elements. A testmerge might be a good idea next weekend so see how it handles crowds during transitions. But for me I just get a bit of a hard pause.

Shuffl3:
Looks like this was put in a week or two before he started the refactor.

Shuffl3:
Im just going to close this out for now and work on bringing in the entire MC/subsystem refactor.

PR #482, opened 2017-02-08
Blatant Thievery

Shuffl3:
Updates heatpipe icons to Tofa01's new set

https://github.com/tgstation/tgstation/pull/23067

🆑Tofa01
Fix: Moved all sprites for heat pipe manifold either up or down by one so that they will line up correctly when connected to adjacent pipes.
/🆑


monster860:
Needs changelogs

PR #483, opened 2017-02-08
Job system first

Shuffl3:
Users complain and get frustrated when they have nothing interactive to dick with during initialization.

This moves the jobs subsystem to the front of the queue so they can nitpick with something instead of contemplating life outside or working out their insecurities in chat.

/tg/ did it so it should be fine right?

https://github.com/tgstation/tgstation/pull/23730


Vivalas:
👍

PR #486, opened 2017-02-11
Fixes pirate ships not fleeing properly

Vivalas:
:cl: Vivalas

fix: Pirate ships should now actually flee.

/:cl:


Qustinnus:
they're pussies now.

PR #491, opened 2017-02-12
Ship Systems Power Management Console

Injazz:
Revamps the power management console which makes it actually working.
You now can manage the power consumption of the important ship systems - currently supported: 

- FTL Drive
- Shield Generator
- Phase Cannons

More power consumption - faster recharging rate.

Preview of the console screen:
image

:cl: injazz
tweak: power management console now manages the power consumption between ship systems
/:cl:


Injazz:
image
and that's how it looks like when everything is charged

PR #511, opened 2017-02-18
Changes the 'station' to 'ship' in the charter announcement and other places

Funce:
So I'm sure this needs to be fixed. I should probably fix the rest of the 'station' but these were the most pressing.
Make sure to check that the description changes are okay.

:cl: Funce
fix: The Charter now refers to the ship as a ship rather than a station.
/:cl:

### DIFF
--- a/code/game/objects/items/charter.dm
+++ b/code/game/objects/items/charter.dm
@@ -3,7 +3,7 @@
 	icon = 'icons/obj/wizard.dmi'
 	icon_state = "scroll2"
 	desc = "An official document entrusting the governance of the ship \
-		to the Captain."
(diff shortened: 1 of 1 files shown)


code/game/objects/items/charter.dm
```diff
@@ -2,8 +2,8 @@
 	name = "station charter"
 	icon = 'icons/obj/wizard.dmi'
 	icon_state = "scroll2"
-	desc = "An official document entrusting the governance of the station \
-		and surrounding space to the Captain. "
+	desc = "An official document entrusting the governance of the ship \
+		to the Captain."
```
monster860: Ship and the area within it's shield boundaries?

PR #512, opened 2017-02-18
Removes all instances of 'station' in the AI modules

Funce:
The AI now recognises the NSV Astreaus as a ship.
Special note to check the ProtectShip module as I don't know if anything specifically uses that value.
My reasoning is its an item and passed.

:cl: Funce
fix: All AI modules now refer to the ship as a ship.
/:cl:


Funce:
Hur dur,remove the ship into the station. NOT

PR #526, opened 2017-02-20
Constructable library

ninjanomnom:
@Ecbrd I gotchu fam

Makes the book binder and scanner able to be constructed via circuit boards

:cl: ninjanomnom
add: Can now print woody's got wood
/:cl:


ninjanomnom:
@monster860 This should be good to merge whenever. Unless we don't want even the possibility of books?

PR #530, opened 2017-02-21
Cyborg balancing

ninjanomnom:
Ion pulse thruster is 3 energy rather than 5

Only use power once every ~~5~~ 3 ticks, effectively reducing power draw. (I couldn't just decrease the amount of power used normally because it was already using only 1 charge by default per tick)

Self repair is 1/3 as effective

(Watch in horror as I mangle github from not using it in ages)

:cl: ninjanomnom
tweak: Cyborg self repair has been nerfed
tweak: Cyborgs use less power
/:cl:


ninjanomnom:
@Qustinnus changes to the self repair nerf and power usage reduction are easy if you're not satisfied with what's here.

ninjanomnom:
Did a bit of testing and found the first pass was quite a bit overtuned, the power cell would last for 2-3 hours

Qustinnus:
@monster860 ITS NERFED YES

PR #532, opened 2017-02-24
Library stuff (again, sorry monster)

ninjanomnom:
Makes the library scanner and binder able to be constructed from circuit boards from science.

:cl:ninjanomnom
add: Can now print woody's got wood
/:cl:


Ecbrd:
<3

PR #566, opened 2017-03-02
AI Ship message temporary fix

### LINKED ISSUE #103: No "combat text" for AI player when in ship to ship combat.
**Problem Description**:
No hit/miss info when playing as the AI.

Why do you think this is a bug?

**What happened instead**:
How is what happened different from what you expected?

Why do you think this is an important issue?

The most important section. Review everything you did leading up to causing the issue.

anything else out of the ordinary happening?)**: Anything else you can tell us.

defrost256:
Fix #103 No "combat text" for AI
Temporary because no custom message format, and no filtering (so floods the AI player with a lot of ships jumping around in the sector)

### DIFF
--- a/code/controllers/subsystem/ships.dm
+++ b/code/controllers/subsystem/ships.dm
@@ -276,6 +276,8 @@ var/global/list/ftl_weapons_consoles = list()
 		return //don't need information about every combat sequence happening across the galaxy
 	for(var/obj/machinery/computer/ftl_weapons/C in ftl_weapons_consoles)
 		C.status_update(message,sound)
+	for (var/mob/living/silicon/aiPlayer in player_list)
+		aiPlayer << message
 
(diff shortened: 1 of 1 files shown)


defrost256:
Alright, just to put it out here, I made this PR mostly to raise awareness to the problem at hand, and the piece of code most likely to contain the solution.
IMO there are two ways of making this happen:

1. By adding some filtering to the messages based on `var/datum/starship/S`, and also a lot of formatting to make it more distinguishable. (which might, or might not imply changing the format of the string it gets)

1. By adding a `weapon/obj/machinery/computer/ftl_weapons/` console right next to the AI so they can hear it.

Whichever you find best suited is up to you, but I can only implement the first one, cause I have no idea how to edit the map.

Qustinnus:
@Vivalas

Vivalas:
I honestly like this better than adding a console to the AI core. :+1:

PR #585, opened 2017-03-23
Library deconstruction fix

ninjanomnom:
:cl: ninjanomnom
fix: Can now break the librarian's shit when he reads WGW on the radio
/:cl:

### DIFF
--- a/code/modules/library/lib_machines.dm
+++ b/code/modules/library/lib_machines.dm
@@ -452,12 +452,16 @@ var/global/list/datum/cachedbook/cachedbooks // List of our cached book datums
 	anchored = 1
 	density = 1
 	var/obj/item/weapon/book/cache		// Last scanned book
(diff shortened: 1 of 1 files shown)


code/modules/library/lib_machines.dm
```diff
@@ -452,25 +452,47 @@ var/global/list/datum/cachedbook/cachedbooks // List of our cached book datums
 	anchored = 1
 	density = 1
 	var/obj/item/weapon/book/cache		// Last scanned book
+	var/stage = 0
```
CrAzYPiLoT-SS13: May I suggest a bit more descriptive of a name, such as 'hatch_opened'?

ninjanomnom: this is the same variable name as other machine deconstruction procedures, that's why i did it this way

PR #590, opened 2017-03-29
Makes minimaps work

Shadowlight213:
Last time when I ported my refactor, I forgot to update the map path procs.

I also changed the minimap to handle the ship being on alternate zlevels.
getflaticon() is also removed, so minimap generation shouldn't cause crashes.

:cl:
fix: Crew Monitor minimaps should actually work now
/:cl:


Vivalas:
@monster860

PR #592, opened 2017-03-31
Makes ftl and shields eject non-plasma gases and minor performance increase

ninjanomnom:
Same as title, I also cached the gases since that's how the atmos code elsewhere says to do it for performance.

:cl: ninjanomnom
fix: Makes non-plasma gases eject from the ftl and shielding machinery
/:cl:


ninjanomnom:
@monster860 speedmerge this shit

PR #576, opened 2017-03-17
Fixes planet gravity and lighting

monster860:
:cl: monster860
fix: Fix planet gravity being broken when the ship leaves
fix: Fix space tiles turning pitch black
/:cl:

PR #593, opened 2017-03-31
Revert "Fixes planet gravity and lighting"

ike709:
Reverts FTL13/FTL13#576

When you land on a planet, ship lighting breaks.


ike709:
Monster potentially fixed this in a commit.

PR #599, opened 2017-04-01
Ports playtime tracking from Paradise

ike709:
:cl: ike709
add: Playtime is now tracked. Only viewable by admins.
/:cl:

**_Insert usual generic disclaimer_**

Ports playtime tracking from Paradise since @Qustinnus is too lazy to do it themself.

It divides playtime into five categories: Living, Ghost, Crewmember, Special, Exempt. Exempt will be needed if someone _does_ add job restrictions based on playtime. Special, I assume, is special stuff like deathsquad.

Original PR: https://github.com/ParadiseSS13/Paradise/pull/5224 Credit to Kyep.

Things to note:
- This doesn't include restricting jobs based on playtime, since I don't like that idea. You can add that yourself if you want.
- I've tested this even less than usual because I don't even know what SQL is aside from _something something_ database.
- It is disabled by default in the config, so you'll have to enable it.
- The original PR includes some post-merge install instructions, only the first one should be relevant.
- You should probably just read the entirety of the original PR, honestly.
- You should actually test this because there's a 50/50 chance it'll break something or just not work at all.


Qustinnus:
idek how to port stuff please teach

Vivalas:
@monster860 You need to review this since it messes with SQL and may burn horribly or something.

ike709:
This almost certainly will burn horribly or just not work at all.

Qustinnus:
@monster860

ike709:
Conflicts resolved.

ike709:
>Your code

nice joke

But yeah in the month since I _completely ported this code and didn't write any of it_ I think I've learned enough to save it in a savefile rather than SQL. I'll work on it sometime this week.

ike709:
I surrender.

PR #606, opened 2017-04-02
Speeds up icon smoothing, hopefully

ike709:
:cl: ike709
tweak: Icon smoothing should be faster, hopefully. Probably broke everything.
/:cl:

Some staffmember said icon smoothing was the thing causing lag, so here you go. This is pretty much the best we can do until overlay queueing and another PR are ported, the latter of which I'm working on.

_**Insert generic disclaimer here**_

Also some stuff uses a custom smoothing proc, and I left it alone because I don't understand it.


ike709:
This will be included in another PR I'm currently working on.

PR #614, opened 2017-04-04
Gibber no longer produces an absurd amount of meat.

### LINKED ISSUE #575: Gibber in chef's kitchen produces an extremely large amount of gibs and meat
monkey gibs

**Problem Description**:
What is the problem?

When gibbing something like a monkey in the gibber in the kitchen, a lot goes wrong. (I have not tested this for human corpses.) First off, there is about 40 pieces of meat produced. Secondly, there is an equal amount of gibs to go with the meat. There is, as you can see in the screenshot, a full list of gibs in a single tile. There is even more in places where you can't see it and can't clean it, such as under the steel tables where the chef puts food and supplies on. It's extremely laggy and seems to lag the server a lot until all of the gibs finish processing. With multiple corpses, the server could potentially crash the people playing, or maybe even the server.

Why do you think this is a bug?

I expected to get a normal amount of meat from the monkey corpse that I gibbed. Note that the meat is labelled "Monkey II meat" only because the monkey was labelled "monkey II meat." It's the same as gibbing someone called "John Smith" and getting "John Smith meat."

**What happened instead**:
How is what happened different from what you expected?

I got about 40 pieces of meat and hundreds of gibs all over the place.

Why do you think this is an important issue?

Potential client/server crashes.

The most important section. Review everything you did leading up to causing the issue.

Kill a monkey (make sure he is dead. I did this multiple times to test it out to make sure it wasn't just a one time thing, and they were death both times. I don't know if there is a different result if the monkey is alive.) and then insert him in the gibber. Click the gibber only one time to chop the monkey up.

anything else out of the ordinary happening?)**: Anything else you can tell us.

I tested this over three rounds, I don't think it matters based on what round it was. Nothing else out of the ordinary was happening. We were not in FTL travel nor were we preparing for it.

Funce:
This fixes #575 

I crashed my test server whilst testing this. I managed to get the gibber to produce over 600 pieces of meat. NOT GOOD FRIENDS.

@monster860 

:cl: Funce
fix: The kitchen's Gibber no longer produces an absurd amount of meat. (Enough to break the server)
/:cl:

### DIFF
--- a/code/modules/food_and_drinks/kitchen_machinery/gibber.dm
+++ b/code/modules/food_and_drinks/kitchen_machinery/gibber.dm
@@ -61,6 +61,7 @@
 
 /obj/machinery/gibber/RefreshParts()
 	var/gib_time = 40
+	meat_produced = 0
 	for(var/obj/item/weapon/stock_parts/matter_bin/B in component_parts)
 		meat_produced += 3 * B.rating
 	for(var/obj/item/weapon/stock_parts/manipulator/M in component_parts)


code/modules/food_and_drinks/kitchen_machinery/gibber.dm
```diff
@@ -61,6 +61,7 @@
 
 /obj/machinery/gibber/RefreshParts()
 	var/gib_time = 40
+	meat_produced = 0
```
ninjanomnom: This is doing to meat_produced what gib_time does for gibtime. You may want to handle both those vars in the same way just to keep it clean looking.

Funce: It's weird, But I wouldn't fix what already works. gib_time is a new variable, and eventually transfers to gibtime. I have no IDEA why.

PR #616, opened 2017-04-04
Crafted stunprods can now actually be used

### LINKED ISSUE #578: Stunprod will not take power cells, even the highest capacity ones.
**Problem Description**:
When you put in a power cell of any capacity that you can make from the autolathes, they are not powerful enough for the stunprod to utilize.

I expected to be able to put cells, including a cell known as a 'super capacity cell', into a device in order to be able to power it.

**What happened instead**:
It does not accept the power cell because it is too weak for the stunprod.

It makes making a stun prod completely pointless because there is no way you would be able to use it with the exception of getting an admin to help you with it. 

1. Get a power cell, from an autolathe on the ship. Anything up to super power cell, at least, will show the problem.
2. make a stun prod.
3.attempt to put the power cell into the stun prod
4. Cry as your stunprod refuses it, making it useless

anything else out of the ordinary happening?)**: Nothing else to tell here.

Funce:
This fixes #578 

Another which I thought was infinitesimally hard. But turned out not so much.

Also Balancing is needed for the number.

:cl: Funce
fix: Crafted stunprods can now actually be used. Don't know why you'd use them though...
/:cl:

### DIFF
--- a/code/game/objects/items/weapons/stunbaton.dm
+++ b/code/game/objects/items/weapons/stunbaton.dm
@@ -171,7 +171,7 @@
 	force = 3
 	throwforce = 5
 	stunforce = 5
-	hitcost = 2500
+	hitcost = 500
 	slot_flags = null
 	var/obj/item/device/assembly/igniter/sparkler = 0
 


ninjanomnom:
For reference, the cyborg power use change I did made their power costs 1/3 what it was before.

PR #625, opened 2017-04-08
MAC cannon now faces east when built. Have fun.

ike709:
:cl: ike709
fix: The MAC cannon will now face the correct direction when constructed. Have fun.
/:cl:

I actually tested this and it actually works.


Vivalas:
>Not just making it rotatable.

ike709:
@Vivalas 
image
Also rotate() code is a fucking mess of copypasta.

PR #629, opened 2017-04-08
Temporarily removes "Kill pirates" objective

ike709:
:cl: ike709
tweak: You'll no longer have an objective to kill pirates. For now.
/:cl:

Something is wrong with the objective to kill a certain number of pirates. They're fucking damn near impossible to find even in the white systems, and rounds with this objective will spend literally hours just trying to _find_ pirates. I don't know if pirates are somewhat broken, need to spawn more frequently, or what.

This is one of the primary sources for hella long rounds that I've seen.


Vivalas:
I can probably actually fix this with a quick two lines, don't wanna remove this completely yet.

Vivalas:
#634

PR #630, opened 2017-04-08
Take 2 on fixing the shields and ftl

ninjanomnom:
:cl: ninjanomnom
fix: FTL and shields are maybe fixed, again
/:cl:


ninjanomnom:
Forgot to also edit the ftl shieldgen in the last fix. Problem might be that the gas mixture remove proc has a garbage collect. This should fix the issue if that's the problem.

PR #631, opened 2017-04-09
Makes minimap generation use docking port sizes

Shadowlight213:
Backup minimaps added for aetherwhisp and trailblazer
Also slightly decreases the aetherwhisp docking port height so it fits within the space station docking bounds. I didn't run into any issues with it during the change.

The trailblazer minimap is missing the top left escape pod. I can't do anything about this due to how minimaps are generated and the shape of the ship unfortunately.


_maps/aetherwhisp.dm
```diff
@@ -17,7 +17,7 @@
 		#define FTL_SHIP_DWIDTH 60
 		#define FTL_SHIP_DHEIGHT 6
 		#define FTL_SHIP_WIDTH 109
-		#define FTL_SHIP_HEIGHT 73
+		#define FTL_SHIP_HEIGHT 71
```
monster860: Why?

Vivalas:
>Works fine
>Doesn't merge

PR #632, opened 2017-04-09
Replaces prize pulse rifle in stations with a pulse destroyer

Shadowlight213:
Fixes ghosts seeing the "someone won a pulse rifle" message when actually it just was spawned on the station.


Qustinnus:
This isn't a bug, it's a cheap way of letting admins teleport to stations before they generate.

Qustinnus:
(which means pls dont merge itll make me cry)

PR #637, opened 2017-04-09
Reduces electrical storm and vent clog weight so they happen less

ike709:
:cl: ike709
tweak: Electrical Storm event weight reduced to 25 from 40
tweak: Electrical Storm cannot happen more than 3 times
tweak: Clogged Vents weight reduced to 30 from 35
/:cl:

Electrical storms happen too damn often, and they offer nothing more than a minor annoyance.

Clogged vents aren't _quite_ as bad but still a bit silly.


ike709:
I'm also tempted to slightly reduce Communications Blackout.

PR #655, opened 2017-04-13
Improves testing QoL by shortening default lobby and roundend time

ike709:
:cl: ike709
tweak: The default time for the lobby and roundend countdowns are much shorter. THESE ARE VALUES FOR TESTING. DO NOT USE THEM ON ACTUAL SERVERS.
/:cl:

## NOTE: DO NOT UPDATE THE CONFIG TO THESE VALUES. THESE ARE FOR TESTING PURPOSES ONLY.

I'm tired of having to reduce these manually all the time, and the config already contains at least one testing default value that would fuck up an actual server.

### DIFF
--- a/config/config.txt
+++ b/config/config.txt
@@ -5,9 +5,11 @@
 STATIONNAME Space Station 13
 
 # Lobby time: This is the amount of time between rounds that players have to setup their characters and be ready.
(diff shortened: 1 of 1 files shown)


monster860:
This is pointless you can just use the admin buttons to start now and hard restart

ike709:
That involves me editing the config to make myself an admin @monster860

PR #676, opened 2017-04-17
Chef ert

Kmc2000:
You asked, you got it....

Never joke about meme PRS with me, I LIVE FOR MEME PRS

This adds an ERT chef equipment option, adminbus only of course.

For those moments when you're just hungry but there's no kitchen out here in space...

Also modifies the survival capsule check to allow you to exempt new pod types from it, just set "exempt" to 1 and it will allow it to deploy...

ANYWHERE....

Expect more meme pods in the future

^Figure one: a kitchen pod crashes through the decks and forces itself upon the NSV aetherwhisp with no regard for any hull breaches it may have caused

Additionally, adds in a somewhat redundant food sign, un-used for now

:cl: Kmc2000
add: Added an adminbussable CHEF ert option, use the set equipment option to use.
add: Added a bluespace meatball, adminbus only, this lets you bluespace in a kitchen...

/:cl:


IlyaIU:
literally hell kitchen

PR #681, opened 2017-04-18
Small runtime purge

ninjanomnom:
Kinda brute force fixed some runtimes from the combat update.

Cannot read null.mode (firealarm.dm: 60) **//Dealt with**

Cannot read null.faction (ships.dm: 413) **//Dealt with**

Cannot read null.ftl_time (ship.dm: 429) **//Dealt with**


ninjanomnom:
Mainly this was to just stop the startup runtime spam from firealarms

PR #686, opened 2017-04-19
Battle Bridge Computers: Correct Icon and Naming Scheme Edition

ike709:
:cl: Kmc2000 & Ike709
fix: Tactical and navigation consoles now follow the conventional console naming scheme.
add: Adds secondary tactical and navigation consoles for the battle bridge. They only work during General Quarters.
/:cl:

@Kmc2000 Every time you tried to fix the grammar, you either didn't correctly fix it or didn't fix it on both computers. Your icons were wrong, nav computers use a different keyboard. This will cause conflicts, by the way.

Also changed the names of the normal consoles to match every other console's naming scheme.


Kmc2000:
Oh.

Yeah alright merge this one instead

ike709:
See: https://github.com/FTL13/FTL13/pull/695

PR #687, opened 2017-04-19
Fix the fire alarm runtime spam at startup

monster860:
No need for changelogs since this isn't a player-facing change.


ninjanomnom:
This is contained in #https://github.com/FTL13/FTL13/pull/681

PR #721, opened 2017-04-23
Quick fix for datum/reagents getting it's flag var assigned as null

ninjanomnom:
:cl: ninjanomnom
fix: Reagents should no longer occasionally refuse to react
/:cl:


ninjanomnom:
okay no this isn't it

PR #725, opened 2017-04-23
SM Rebalancing for subsystem change

ninjanomnom:
Because it seems the tickrate is faster on SSair.

Made n2 reaction slowdown higher because it's a shit gas and should have *some* use.
Only runs every other air tick now.
Emitter power gets used slightly faster.

This may be overtuning, I'll test it more when people dont fuck it up every round

:cl: ninjanomnom
tweak: SM Rebalance
/:cl:

### DIFF
--- a/code/modules/power/supermatter/supermatter.dm
+++ b/code/modules/power/supermatter/supermatter.dm
@@ -2,7 +2,7 @@
 //Please do not bother them with bugs from this port, however, as it has been modified quite a bit.
 //Modifications include removing the world-ending full supermatter variation, and leaving only the shard.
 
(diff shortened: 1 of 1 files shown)


code/modules/power/supermatter/supermatter.dm
```diff
@@ -88,6 +90,12 @@
 	qdel(src)
 
 /obj/machinery/power/supermatter_shard/process_atmos()
+	if(tick_count == tick_interval && tick_interval != 0) //If tick_interval is 0 just ignore this block
```
Vivalas: What does this block even do?

Vivalas: Never mind I see the return,

PR #735, opened 2017-04-25
Boarding fixes and tweaks

IlyaIU:
Fixed a mess happened on a first public test

:cl: Injazz
tweak: Adjusted Self-Destruct timer to 12 minutes.
tweak: Announcment for boarding event happens as soon as the ship got hit.
tweak: Self Destruct Device now anchored.
fix: fixed a map issues
fix: fixed an incorrect timer display
fix: fixed event not being called only one time in a round
/:cl:


code/game/gamemodes/miniantags/ftl/enemy_ship/boarding.dm
```diff
@@ -8,12 +8,12 @@
 	var/datum/planet/planet = null
 	var/shield_down
 	var/time_set
-	var/timer = 420 //time before event end (7 minutes)
+	var/timer = 720 //time before event end (7 minutes)
```
ike709: Update the comment.

IlyaIU: That's still 7 minutes. Consider that as a fix, since defender got 5 minutes to prepare before shields go down, so technically there is only 7 minutes for event remaining

Cirrial: Might be worth clarifying the comment for future reference, then.

IlyaIU: nah

PR #737, opened 2017-04-26
Bugfix for #707 - fix runtime on item_attack.dm related to missing attack verbs

### LINKED ISSUE #707: Runtime in item_attack.dm, line 90: Cannot read null.len
image

Cirrial:
Fixes #707. The attack_verb list might be null, so check for its existence first. Simple enough.

### DIFF
--- a/code/_onclick/item_attack.dm
+++ b/code/_onclick/item_attack.dm
@@ -87,7 +87,7 @@
 
 /mob/living/proc/send_item_attack_message(obj/item/I, mob/living/user, hit_area)
 	var/message_verb = "attacked"
-	if(I.attack_verb.len)
+	if(I.attack_verb && I.attack_verb.len)
 		message_verb = "[pick(I.attack_verb)]"
 	else if(!I.force)
 		return 0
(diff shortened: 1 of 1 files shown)


code/_onclick/item_attack.dm
```diff
@@ -110,4 +110,4 @@
 		visible_message("<span class='warning'>[I] bounces harmlessly off of [src].</span>",\
 					"<span class='warning'>[I] bounces harmlessly off of [src]!</span>")
 	else
-		return ..()
\ No newline at end of file
+		return ..()
```
monster860: Uhhh.....

Cirrial: If we had standardised line endings this sort of thing wouldn't happen! I can't fix it without introducing MORE line ending problems at the moment, so I guess I'll close this and try again when my repository is less on fire.

monster860: Try text editing using Atom. http://atom.io

Cirrial: I've been using Sublime Text and it only just occurred to me to look for a plugin that can solve this problem. Agh!!

PR #770, opened 2017-05-05
Stops shipspam

ninjanomnom:
This shouldn't be needed anymore and it's spamming up admin chat

### DIFF
--- a/code/datums/ship.dm
+++ b/code/datums/ship.dm
@@ -322,6 +322,7 @@ var/next_ship_id
 	if(!istype(chosen_target)) //if "ship" is picked.
 		ship.attacking_player = 1
 		SSship.broadcast_message("<span class=notice>Warning! Enemy ship detected powering up weapons! ([ship.name]) Prepare for combat!</span>",SSship.alert_sound,ship)
+		message_admins("[ship.name] has engaged the players into combat at [ship.system]!")
 	else
 		ship.attacking_target = chosen_target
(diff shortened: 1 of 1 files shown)


Qustinnus:
why though :(

ninjanomnom:
Cause it's spammy and unnecesary, it makes it harder to read what people are saying.

ike709:
We could always just have a special log for NPC ship stuff...

code/datums/ship.dm
```diff
@@ -326,7 +326,6 @@ var/next_ship_id
 		ship.attacking_target = chosen_target
 		SSship.broadcast_message("<span class=notice>Caution! [SSship.faction2prefix(ship)] ship ([ship.name]) locking on to [SSship.faction2prefix(ship.attacking_target)] ship ([ship.attacking_target.name]).</span>",null,ship)
 
-	message_admins("[ship.name] has entered into combat at [ship.system]! [ship.attacking_player ? "" : "Combat was not due to players!"]")
```
monster860: Maybe make it only happen when the combat is triggered by players.

ninjanomnom: I can do that

Vivalas: I agree, the earlier behavior was just to help understand what was going on under the hood with ships due to their abstraction, and to help pinpoint erratic behavior with ships if it happens

ninjanomnom:
https://puu.sh/vI3bM/398a4cfc02.png

Stealthkibbler:
Could you maybe change it so the non player triggered ones show up on the news casters under the galactic news channel? I know viv did that with the economy.

ninjanomnom:
That's overscoped for this pr. This is just to make chat more readable

PR #771, opened 2017-05-05
Adds hydrogen to atmosia

ninjanomnom:
:cl: ninjanomnom
add: Hydrogen is now available in the previously empty storage tank in atmosia.
/:cl:


ninjanomnom:
Aerwhisp is doing this now

PR #772, opened 2017-05-05
The PDA server starts enabled by default now

ike709:
:cl: ike709
tweak: The PDA server starts enabled by default now.
/:cl:


ike709:
@monster860 I did it mom

code/modules/research/message_server.dm
```diff
@@ -102,7 +103,10 @@ var/global/list/obj/machinery/message_server/message_servers = list()
 	//	decryptkey = generateKey()
 	if(active && (stat & (BROKEN|NOPOWER)))
 		active = 0
-		return
+	else if(toggled)
+		active = 1
+	else
```
Vivalas: Won't this make it turn off now as soon as you turn it on then?

ike709: No. I've tested it and it works as intended. The PDA server is initially on, it can be turned off, and it can be turned back on. It works perfectly.

ike709: Changed this to make it easier to read. Still has the same effect.

PR #773, opened 2017-05-05
Makes pipe filter and mixer turn back on after power outage

### LINKED ISSUE #767: Filters still stay off
**Problem Description**:
If there is a power failure the filters turn off, but when the power turns on again the filters stay off.

I expectet the filters to turn back on, and this was fixed before i believe.

**What happened instead**:
The filter didn't turn on.

The filters in the hot loop turn off, stops the gas flow, power generation, and if you're unlucky you can say bye to the sm. And the scrubber network also fails because all he scrubbers are off.

Turn on filters in a working setup, turn off power or wait for a short power outage, turn power back on, look as filters stay off.

anything else out of the ordinary happening?)**: We got hit and that caust a short power failure too.

ninjanomnom:
:cl: ninjanomnom
tweak: Pipe filters and mixers turn back on once they have power if they were on before
/:cl:

Fixes #767

### DIFF
--- a/code/modules/atmospherics/machinery/components/trinary_devices/filter.dm
+++ b/code/modules/atmospherics/machinery/components/trinary_devices/filter.dm
@@ -4,6 +4,7 @@
 	density = 0
 	can_unwrench = 1
 	var/on = 0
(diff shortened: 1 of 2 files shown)


Vivalas:
🙏

PR #782, opened 2017-05-06
SM Chamber vent fix

Raschert0:
:cl:
tweak: Setting target pressure by using Engine Cooling Control now will correctly affect vent in SM chamber. Vent will try siphoning gases to equalize pressure in connected pipe with target pressure.
/:cl:


ninjanomnom:
It would be better if you make it so the console can change the external pressure checks.

PR #784, opened 2017-05-07
Adds Keek's Mesh tile icons

Funce:
Just a side project Keek wanted done, this puts in the icons. They're pretty mediocre, but putting them here allows Keek to put the new ones where they're supposed to.

These will be used in my next PR.

First is the item, the second is the structure.
image
image

:cl: Funce
imageadd: Keek's mesh tiles are now on the codebase for use.
/:cl:


ike709:
Pics or it didn't happen

Funce:
First is the Mesh tiles, the item, then the actual what it looks like.
image
image

Funce:
Added.

Funce:
cheers to ike for pointing it out. These meshes are now good for secure use.

monster860:
Can't you just set the mouse_opacity to 2 in order to prevent clicking through it?

ike709:
@monster860 No. The same thing was suggested on the PR that prevented clicking on grilles, and that idea was shot down. Forget why. Feel free to find the PR on tg though.

Funce:
@monster860 
image

monster860:
Hm... The alpha thing feels like a hack though.

Funce:
Yeah it is a hack but limited probably only to these sorts of things that could change their sprite during a round. If there was another way that would be nice and simple to manage, I would. Otherwise it really isn't that bad. Unless you're worried about something specific.

Vivalas:
Hm, on CM these function like actual turfs and don't show anything under them in even the context box. I'm guessing they use underlays and mess with the invisiblity of the underlying pipes and stuff, not sure.

Funce:
Oh god, if I have to play with that I think I'd literally die. But that would be an implementation issue, rather than anything else UNLESS you want these to be in the Turfs.dmi. Because this is just the asset files for updating purposes.

Funce:
@ninjanomnom Done Done and Done

PR #792, opened 2017-05-07
Add ability to take metal/glass out of a drone shell dispenser

NIXC:
Just a small add, you can now remove the metal and glass. Useful if you fill it with to much metal and glass (9000 drone shells anyone?) or fill it with all of one type so it gets stuck and becomes useless. I can change how its extracted instead of a crowbar if anyone has any better ideas.
You cant get the last half sheet out since its too small.

:cl: NINXE
add: Using a crowbar on a Drone shell dispenser removes stored material
add: Drone dispenser is now construtable and deconstructable
/:cl:


ike709:
Could do a right click verb called "Eject Materials"

NIXC:
Ok now you can build and destroy drone dispensers to your hearts content. screwdriver then crowbar to disasemble, just crowbar to remove material. I did not add anyway to spawn the board

PR #346, opened 2016-12-30
Removes carp migration and immovable rod events

monster860:
:cl: monster860
del: Disable carp migration and immovable rod events
/:cl:

Carp migration happens literally every other round and it sucks, and immovable rod fucks over any round for no reason.

PR #793, opened 2017-05-07
Fix of null.len runtime in ship.dm at line #346

Raschert0:


code/datums/ship.dm
```diff
@@ -343,8 +343,11 @@ var/next_ship_id
 
 	if(ship.target_system && ship.system && !ship.is_jumping)
 //		message_admins("[ship.name] [ship.system] [ship.mission_ai.cname] [ship.target_system] [ship.system]")
-		if(!ship.star_path.len) //dunno how this happens yet but it does
+		if(!ship.star_path || !ship.star_path.len) //dunno how this happens yet but it does
 			ship.star_path = get_path_to_system(ship.system, ship.target_system, ship.ftl_range, 200)
+			if(!ship.star_path.len) //If we still can't make path...
```
ninjanomnom: This also needs !ship.star_path, len can possibly runtime in the worst case

Raschert0: but get_path_to_system will return at least clear list()
And runtimes here is because ship.star_path isn't initialized (or i can't find where it occurs).

ninjanomnom: no reason not to do it though and it removes the possibility of a runtime

Raschert0: *sigh

PR #811, opened 2017-05-15
Replaces Pulser Prizes with Station Landmarks

IndusRobot:

Replaces the gun prizes on both stations with an effect that was made a while back. These should still notify the admins that the station has loaded. Ike and I have been having issues with map merging for a while, so I gave up and pushed anyway. 

:cl: IndusRobot
add: Replaces the pulser prizes in station giftshops with station object effects that notify admins. 
/:cl:


ike709:
Multiple people have tried and failed to do this small change without a massive number of line changes.

I don't know if it's line endings or what, but editing these two maps in fastdmm results in this massive diff.

But this is okay for two reasons:
1. Nobody ever touches these files, so conflicts are irrelevant.
2. This initial conversion or whatever is happening will probably not need to be done again.

ninjanomnom (approved):
The large file change is likely because it wasn't in the correct format in the first place, someone made it in dm.

PR #827, opened 2017-05-21
[Aetherwhisp] Brand new Munitions layout

IndusRobot:

A brilliant idea came to me while I was trying to figure this problem out. How about, nudge everything forward!

dreamseeker_2017-05-20_19-24-59

(Second MAC not included.)

Moves the machine frame in munitions forward one tile
Replaces one reinforced wall with reinforced floor tile. 
Cleans up some colored floor tiles. 

:cl: optional name here
fix: [Aetherwhisp] Moves the machine frame in Munitions forward, so an MO doesn't accidentally block the MAC after construction. 
/:cl:


ninjanomnom (approved):
"completely new"

PR #828, opened 2017-05-21
Adds books for SOP and Department SOP, and mapped them in.

ike709:
:cl: ike709
add: Added books for Standard Operating Procedure and Department SOPs. Every department and head of staff gets a copy.
tweak: Renamed Space Law to Ship Law in compliance with the wiki.
/:cl:

Merge ready.

Renamed Space Law to Ship Law like it is on the wiki.
Added books for SOP and every department SOP, but I need a sprite. Currently uses a generic book icon.
Every department gets one copy of their SOP, and every head of staff gets a copy in their office.
Command gets two because they're special. They also get two copies of the main SOP.
I only mapped the books in on Aetherwhisp though.


ike709:
Added a my own coder sprite. It's not great but it works. @ninjanomnom 
image

PR #831, opened 2017-05-23
Chemistry Jazz

Shuffl3:
Fixed the spray bottle's list for reagent transfers so it'll work in the chem dispenser
Corrected the chem masters description for the maximum volume of a patch

### DIFF
--- a/config/jobs.txt
+++ b/config/jobs.txt
@@ -31,7 +31,7 @@ Ship Engineer=5,5
 Atmospheric Technician=3,2
 
 Medical Doctor=5,3
-Chemist=2,2
+Chemist=1,1
 Geneticist=2,2
 Virologist=1,1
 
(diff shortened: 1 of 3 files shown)


Shuffl3:
Guess im gonna try to limit the amount of roundstart chemists again too

config/jobs.txt
```diff
@@ -31,7 +31,7 @@ Ship Engineer=5,5
 Atmospheric Technician=3,2
 
 Medical Doctor=5,3
-Chemist=2,2
+Chemist=1,1
```
ike709: Config files won't be updated on the server, they need to be edited manually.

@ninjanomnom @Vivalas @monster860

ninjanomnom:
I've updated the job config, you can remove that change from this pr

PR #832, opened 2017-05-25
Adds alphabetizer tool to tools folder

NIXC:
Everything of interest should be covered in the read me or the (shudders) comments.
Ill use it on the FTL icons when the rebase is over, if we do it now it will just make merge conflicts all over the place and I have no wish to fix that.

Any suggestions on the code feel free, I'm all ears.


ninjanomnom:
I'm going to wait and see what the people over at TG have to say about your pr there. I don't know exactly how this will affect things if at all.

NIXC:
Well the actual alphabetization is dead, but this is somewhat handy as a jumping off point for related dmi projects. I leave it in your hands weather to merge or close.

ninjanomnom:
If TG doesn't alphabetize, it's too much effort to maintain different dmis than them. And since we won't alphabetize I see no reason to include this in the tools folder.

PR #836, opened 2017-05-31
Mirrorbot sh

ninjanomnom:
A needed file for the mirrorbot


ninjanomnom:
Actualy, I'll just include the needed file in the rebase

PR #838, opened 2017-06-02
Moves items from mining shuttle corners

ike709:
The same stuff is still there, just in different places within the same rooms. We did lose one poster though.


ike709:
I'm a dumbass

PR #945, opened 2017-06-09
Goonchat!

ike709:
🆑 Lzimann + Cyberboss
experiment: Ported goonchat. Much less laggy and crashy than BYOND chat. + Frills!
/🆑

@Vivalas Was that so hard?

Closes #933


Vivalas:
I never implied that it would be

PR #973, opened 2017-06-10
Phase cannon nerf

ninjanomnom:
:cl: ninjanomnom
balance: Increased firing cost of phase cannon by 2,5x
/:cl:

Because rebase


ftl13.dme
```diff
@@ -241,9 +241,9 @@
 #include "code\datums\ship.dm"
 #include "code\datums\shuttles.dm"
 #include "code\datums\soullink.dm"
-#include "code\datums\verbs.dm"
 #include "code\datums\star_economy.dm"
 #include "code\datums\stars.dm"
+#include "code\datums\verbs.dm"
```
monster860: oi you pleb you're using dream maker. use atom instead.

ninjanomnom: dream maker reorders the `#include` lines when you compile, that doesn't mean I edit in it.

PR #980, opened 2017-06-10
Ports planet floors to initialize

### LINKED ISSUE #963: Planetside lighting is broken again
It's all darkness

AuroranAI:
Fixes #963
Planet edges are still fucked, but this makes it playable.


Vivalas:
Ah, this must also fix those assimilate runtimes since the planet tiles will actually have `air`

PR #986, opened 2017-06-10
organizes the shuttleMove procs a bit

ninjanomnom:
no front end changes here, just a bit of cleanup from the rebase

edit: pulling along stationary docks might be a cause of some bugs if any code checks if the ship is docked

### DIFF
--- a/code/modules/shuttle/shuttle.dm
+++ b/code/modules/shuttle/shuttle.dm
@@ -21,8 +21,6 @@
 	var/dwidth = 0	//position relative to covered area, perpendicular to dir
 	var/dheight = 0	//position relative to covered area, parallel to dir
 
-	shuttle_abstract_movable = 1
-
 	//these objects are indestructible
 /obj/docking_port/Destroy(force)
 	// unless you assert that you know what you're doing. Horrible things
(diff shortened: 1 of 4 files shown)


code/modules/shuttle/shuttle.dm
```diff
@@ -21,8 +21,6 @@
 	var/dwidth = 0	//position relative to covered area, perpendicular to dir
 	var/dheight = 0	//position relative to covered area, parallel to dir
 
-	shuttle_abstract_movable = 1
```
Vivalas: Wut.

ninjanomnom: You dont need this here, this would make it so stationary docks move. You added shuttle_abstract_movable to mobile ports already like I asked.

ninjanomnom: Which now that I think about it, it should be the other way around, I forget if I told you to add it to all docks or only mobile docks during the rebase but you aparently added it to both.

Vivalas: I thought I had just added it to mobile

ninjanomnom: Wait no yeah this is correct, stationary ports on the ship have to have the var changed in mapping.

PR #999, opened 2017-06-11
Fix related CID/IPs on player panel

ike709:
Closes #954


ninjanomnom:
this is just exchanging fixes, our codebase fixed this already

ike709:
If you say so...?

PR #1025, opened 2017-06-12
Fixes gravity

### LINKED ISSUE #965: No gravity during FTL jumps
AHHHHHH

AuroranAI:
Grav gen actually updates now, and deletes itself properly after leaving a z level
fixes #965

### DIFF
--- a/code/modules/power/gravitygenerator.dm
+++ b/code/modules/power/gravitygenerator.dm
@@ -52,15 +52,6 @@ GLOBAL_LIST_EMPTY(gravity_generators) // We will keep track of this by adding ne
 /obj/machinery/gravity_generator/shuttleRotate()
 	return
 
-// No free riding on my our gravity generator. Get your own, you freeloading martians!
-/obj/machinery/gravity_generator/main/beforeShuttleMove()
-	var/turf/T = get_turf(src.loc)
-	if(T)
-		GLOB.gravity_generators["[T.z]"]-= src
-
-/obj/machinery/gravity_generator/main/afterShuttleMove()
-	update_list()
-
 /obj/machinery/gravity_generator/proc/get_status()
 	return "off"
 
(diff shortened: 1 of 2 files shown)


ike709:
You need to add Fixes #965

code/modules/power/gravitygenerator.dm
```diff
@@ -52,6 +52,15 @@ GLOBAL_LIST_EMPTY(gravity_generators) // We will keep track of this by adding ne
...
 
+// No free riding on my our gravity generator. Get your own, you freeloading martians!
+/obj/machinery/gravity_generator/main/beforeShuttleMove()
+	var/turf/T = get_turf(src.loc)
+	if(T)
+		GLOB.gravity_generators["[T.z]"]-= src
+
+/obj/machinery/gravity_generator/main/afterShuttleMove()
+	update_list()
+
```
ninjanomnom: This needs to be in on_move.dm

Also, update_list should be used in both procs. If the gravity gen is off when the ship moves a runtime will happen with current implementation.

code/modules/shuttle/on_move.dm
```diff
@@ -93,6 +93,17 @@ All ShuttleMove procs go here
 	if(z == ZLEVEL_MINING) //Avoids double logging and landing on other Z-levels due to badminnery
 		SSblackbox.add_details("colonies_dropped", "[x]|[y]|[z]") //Number of times a base has been dropped!
 
+/obj/machinery/gravity_generator/main/beforeShuttleMove()
+	update_list() //make sure everything's correct before running
+	if(!on)
+		return
```
ninjanomnom: update list does the same thing as the other code here, you just need to change the `on` var and then update_list()

in the `afterShuttleMove()` you can check charging_state for if it was on before, `if(charging_state == POWER_IDLE)` then set on to true.

PR #1038, opened 2017-06-13
Shuttle lighting fix

### LINKED ISSUE #1024: No lighting during FTL jumps
After the first jump, lighting will no longer work during FTL jumps.

ninjanomnom:
Spent all day looking at shuttle code again and had to track this bug down from ground zero by comparing turf vars before during and after transit. If anyone breaks this again and I have to fix it a third time they can go fix it themselves.

I also deleted an unused proc in shuttle.dm

Fixes #1024 

:cl: ninjanomnom
fix: Lights are fixed in the second jump onward now
/:cl:


ike709:
Fixes #1024

PR #1042, opened 2017-06-13
Revert "Fixes the readme and makes it more accurate"

ninjanomnom:
But leaves the fix in.

@Vivalas merged it without any communication even though it's a subjective change.

Also too many badges is gross.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,8 @@
-## FTL13
+##FTL13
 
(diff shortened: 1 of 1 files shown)


Vivalas:
@monster860 Reeee maintainer drama.

Would it be better if they were just on newlines?

ike709:
Newlines is worse then you're making it longer for memes

Same line utilizes existing space.

optimumtact:
built with resentment is all you need

PR #1062, opened 2017-06-14
Fixes command reports being named wrong and triggering twice.

Vivalas:
This was me being stupid.


code/modules/admin/verbs/randomverbs.dm
```diff
@@ -517,10 +492,24 @@ Traitors and the like can also be revived with the previous role mostly intact.
...
 
-	print_command_report(crBody,"[confirm=="Public" ? "" : "Classified "][crSender] Report")
+		print_command_report(crBody,"[confirm=="Public" ? "" : "Classified "][crSender] Report")
 
 	log_admin("[key_name(src)] has created a command report: [crBody]")
 	message_admins("[key_name_admin(src)] has created a command report")
+
+/client/proc/cmd_change_command_name()
+	set category = "Special Verbs"
+	set name = "Change Command Name"
```
ninjanomnom: both their names are "Change Command Name"

PR #1098, opened 2017-06-16
Some shuttlecode modernization and fixes ai sight

### LINKED ISSUE #1079: AI vision bug
[Note]: # (Please enter the commit hash and active testmerge numbers from the "Show Server Revision" verb if you can. If you believe the issue to be caused by a testmerge, please report it in its relative PR thread. State what the issue is from a "whats wrong" prospective. Issue reports should clearly allow maintainers to understand whats wrong and how to test/reproduce if that is not obvious. Avoid ambiguity. Start your issue report below both of these lines (or remove them))

[Admins]: # (If you are reporting a bug that occured AFTER you used varedit/admin buttons to alter an object out of normal operating conditions, please verify that you can re-create the bug without the varedit usage/admin buttons before reporting the issue.)

646759
The vision of the AI (from my end, at least) starts perfectly when docked, and then cuts to static in seemingly random locations of the ship, usually including massive chunks of the ship. These areas appear identical to all areas that the AI does not have camera access to; Floors and plating.
The only thing I can tell is that being in orbit causes it, and until it broke, being in FTL (sometimes, somewhat) would fix it.

Initially, this is the view I had for most of the northern part of the station. 

For a while, restarting my client would at least change where I can see. After a while it stopped working, and this is an example of what I was left with. 

 Another example of how strange and random this seems to be. 

Final example
As a note, this may be caused by the lighting subsystem not loading or having errors.

ninjanomnom:
:cl: ninjanomnom
fix: AI sight works now after jumping
/:cl:

Fixes #1079

### DIFF
--- a/code/controllers/subsystem/shuttle.dm
+++ b/code/controllers/subsystem/shuttle.dm
@@ -74,6 +74,7 @@ SUBSYSTEM_DEF(shuttle)
 			if(!P.station)
 				continue
 			P.station.generate()
+			CHECK_TICK
 
 	setup_transit_zone()
 	initial_move()
--- a/code/controllers/subsystem/mapping.dm
+++ b/code/controllers/subsystem/mapping.dm
@@ -161,6 +161,8 @@ SUBSYSTEM_DEF(mapping)
 								qdel(A)
 						for(var/A in T.contents)
 							qdel(A) // Some qdels dump their shit on the ground.
+						if(GLOB.cameranet.chunkGenerated(T.x, T.y, T.z))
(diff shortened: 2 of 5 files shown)


code/controllers/subsystem/shuttle.dm
```diff
@@ -74,6 +74,7 @@ SUBSYSTEM_DEF(shuttle)
 			if(!P.station)
 				continue
 			P.station.generate()
+			CHECK_TICK
```
monster860: Thereee, now we can use watchdog again.

PR #1124, opened 2017-06-17
Fixes lighting on planets and hopefuly all other future fulldark tile bugs.

### LINKED ISSUE #1064: Lighting breaks when landing on planets.

ninjanomnom:
:cl: ninjanomnom
fix: Fixed planet lighting
experiment: Hopefully also fixes all future full dark tile bugs
/:cl:

image

R.I.P my sanity

Fixes #1064


ninjanomnom:
Fixing planet lighting as well real quick

ninjanomnom:
This moves the lighting setup into the lighting subsystem since it was only used there anyway and was cluttering the global namespace. Also added a per z level initialization for when maps are loaded during play.

PR #1164, opened 2017-06-20
Band-aid for landmark fuckup

outbools:
:cl: Injazz
fix: Lootdrop from salvage now spawns correctly
/:cl:


ninjanomnom:
reverting to old worse methods to fix something is shit. Plus this is fixed in #1167

PR #1171, opened 2017-06-20
Boarding z-level deletion and transition

outbools:
:cl: Injazz
fix: boarding z-level now deletes and deallocating when not needed
/:cl:

### DIFF
--- a/code/controllers/subsystem/mapping.dm
+++ b/code/controllers/subsystem/mapping.dm
@@ -234,8 +234,40 @@ SUBSYSTEM_DEF(mapping)
 	else
 		world.log << "Unable to load z-level [PL.z_levels[PL.map_names.len+1]] for [PL.name]! File: [map_name.map_name]"
 	CHECK_TICK
+	GLOB.space_manager.do_transition_setup()
 	repopulate_sorted_areas()
 	PL.map_names += load_name
+	return PL.map_names.len+1
+
+/datum/controller/subsystem/mapping/proc/del_z_from_planet(var/datum/planet/PL, var/chosen_z)
+	log_world("Unloading z-level #[chosen_z] from [PL.name]...")
+	if(!PL)
+		log_world("Planet not found")
+		return
+	if(!chosen_z)
+		log_world("Z-level not found")
+		return
+	var/z_level = PL.z_levels[chosen_z]
+	for(var/datum/sub_turf_block/STB in split_block(locate(1, 1, z_level), locate(255, 255, z_level)))
+		for(var/turf/T in STB.return_list())
+			for(var/A in T.contents)
+				if(istype(A, /obj/docking_port))
+					qdel(A, 1) // Clear everything out. Including docking ports.
+				else
+					qdel(A)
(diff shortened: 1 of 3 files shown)


bridge.dmm
```diff
@@ -1,49 +0,0 @@
-"a" = (/turf/open/space,/area/space)
```
ninjanomnom: lol you forgot to make a new branch

code/controllers/subsystem/mapping.dm
```diff
@@ -205,8 +205,40 @@ SUBSYSTEM_DEF(mapping)
 	else
 		world.log << "Unable to load z-level [PL.z_levels[PL.map_names.len+1]] for [PL.name]! File: [map_name.map_name]"
 	CHECK_TICK
+	GLOB.space_manager.do_transition_setup()
 	repopulate_sorted_areas()
 	PL.map_names += load_name
+	return PL.map_names.len+1
```
ninjanomnom: you don't need to add 1 to len since you just added to the list the line above.

code/controllers/subsystem/mapping.dm
```diff
@@ -205,8 +205,40 @@ SUBSYSTEM_DEF(mapping)
...
+				GLOB.cameranet.chunks -= GLOB.cameranet.getCameraChunk(T.x, T.y, T.z)
+			SSair.remove_from_active(T)
+			CHECK_TICK
+			z_level_alloc -= "[z_level]"
+			z_level_to_planet_loader -= "[z_level]"
+			free_zlevels["[z_level]"] = z_level
+			PL.z_levels -= z_level
+			GLOB.space_manager.do_transition_setup()
+			repopulate_sorted_areas()
+			log_world("Z-level [z_level] for [PL.name] unloaded")
```
ninjanomnom: why is all this in the turf loop

outbools: oh my ATOM broke

bridge.dmm
outbools: that's the same commit so it's gonna be ignored if previous PR got merged and vice versa

outbools: yeah i checked they share same hash

code/controllers/subsystem/mapping.dm
```diff
@@ -205,8 +205,40 @@ SUBSYSTEM_DEF(mapping)
...
+				GLOB.cameranet.chunks -= GLOB.cameranet.getCameraChunk(T.x, T.y, T.z)
+			SSair.remove_from_active(T)
+			CHECK_TICK
+			z_level_alloc -= "[z_level]"
+			z_level_to_planet_loader -= "[z_level]"
+			free_zlevels["[z_level]"] = z_level
+			PL.z_levels -= z_level
+			GLOB.space_manager.do_transition_setup()
+			repopulate_sorted_areas()
+	log_world("Z-level [z_level] for [PL.name] unloaded")
```
ninjanomnom: you only indented log_world, you shouldnt be doing anything under the CHECK_TICK every for loop

PR #1196, opened 2017-06-20
Adds examine tooltips

ike709:
Closes #1151 

🆑 Tacolizard
add: Repurposed the action button tooltip code to add tooltips with the examine details of all held and equipped items. Hover your mouse over any item equipped, held or in your inventory to examine it.
add: Examine tooltips can be toggled with the 'toggle-examine-tooltips' verb, which can be typed or accessed from the OOC menu.
add: You can change the delay before a tooltip appears with the 'set-examine-tooltip-delay' verb, also accessible via the OOC menu. The default delay is 500ms.
/🆑


ike709:
@ninjanomnom boi

PR #1236, opened 2017-06-23
Dressers PR + code cleanup

AuroranAI:
Closes #1212
also turns all of sheet_types one_per_turf / on_floor to use TRUE instead of 1 to keep us in line with upstream.
🆑
add: Dressers are now constructable, and unanchorable with a wrench.
/🆑

### DIFF
--- a/code/game/objects/structures/dresser.dm
+++ b/code/game/objects/structures/dresser.dm
@@ -7,26 +7,16 @@
 	anchored = 1
 
 /obj/structure/dresser/attackby(obj/item/P, mob/user, params)
-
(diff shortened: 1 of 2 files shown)


code/game/objects/structures/dresser.dm
```diff
@@ -6,6 +6,28 @@
...
+		anchored = !anchored
+
+
+
+/obj/structure/dresser/deconstruct(disassembled = TRUE)
+
+  new /obj/item/stack/sheet/mineral/wood (get_turf(src), 10)
+
+  qdel(src)
+
```
ninjanomnom: You seem to have a whites[ace issue here

PR #1358, opened 2017-07-01
MPRF: Stamina projectiles no longer set off grenades and flamethrowers

Kmsxkuse:
Closes #1228


Kmsxkuse:
Strange, why is Github counting identical segments of code changes?

PR #1395, opened 2017-07-02
MPRF: Fix modular consoles

Kmsxkuse:
Closes #1285

🆑 drline
fix: Nanotrasen finally sent out IT personnel to plug your modular consoles back in. You're welcome.
/🆑


Kmsxkuse:
Yes but this is only a MPRF, the change log is found in the original MPR.

PR #1400, opened 2017-07-02
Fixes shuttle init freeze and reimplements ast integration

monster860:
Fun shit


code/controllers/subsystem/ast.dm
```diff
@@ -0,0 +1,17 @@
+SUBSYSTEM_DEF(ast)
```
ninjanomnom: image

Doesn't matter I guess, just one more thing to rewrite.

PR #1477, opened 2017-07-07
[FINISHED] Mentorship

Kmsxkuse:
:cl: Kmsxkuse
Add: Now you can be a mentor, an admin without the fun tools.
/:cl:

THIS IS A WORK IN PROGRESS,
90% complete.


ike709:
I don't have time to actually look at the code so I just quickly skimmed it, but Floyd and I both think it'd be better if mentors had their own channel called msay that admins could also see and use.

Kmsxkuse:
Hmmmm, a seperate channel... Okay. That's gonna take some work.

code/modules/admin/admin.dm
```diff
@@ -10,6 +10,15 @@
 
 
 ///////////////////////////////////////////////////////////////////////////////////////////////Panels
+//TMentor
+/proc/is_mentor(client/C)
+	if(!istype(C))
+		return FALSE
+	if(!C.holder)
+		return FALSE
+	if(C.holder.rank.rights == R_MENTOR)
```
ninjanomnom: Isn't rights a bitflag? This should be `if(C.holder.rank.rights & R_MENTOR)` unless your intent here is to check if the client is *only* a mentor.

code/modules/admin/topic.dm
```diff
@@ -1528,7 +1528,7 @@
 		show_player_panel(M)
 
 	else if(href_list["adminplayerobservefollow"])
-		if(!isobserver(usr) && !check_rights(R_ADMIN))
+		if(!isobserver(usr) && !check_rights(R_ADMIN || R_MENTOR))
```
ninjanomnom: `|` not `||`

code/modules/admin/admin.dm
Kmsxkuse: Yea, that's for only mentors. To check if the person is a mentor, not just any admin.

Kmsxkuse: Heck, i dont use it anyways so....

code/__DEFINES/admin.dm
```diff
@@ -35,14 +35,8 @@
 #define R_SOUNDS		2048
 #define R_SPAWN			4096
 #define R_TICKET		8192
-
-#if DM_VERSION > 512
-#error Remove the flag below , its been long enough
-#endif
-//legacy , remove post 512, it was replaced by R_POLL
-#define R_REJUVINATE	2
```
ninjanomnom: This shouldn't be removed yet

config/admin_ranks.txt
```diff
@@ -31,7 +32,8 @@
 # END_KEYWORDS
 
 Admin Observer
-Moderator		= +ADMIN
+Mentor = +MENTOR
```
ninjanomnom: Please indent the same as the other entries

Kmsxkuse: I dont see that on my text editor.

code/__DEFINES/admin.dm
Kmsxkuse: Added it back in.

code/modules/admin/holder2.dm
```diff
@@ -17,6 +17,22 @@ GLOBAL_PROTECT(admin_datums)
 	var/datum/newscaster/feed_channel/admincaster_feed_channel = new /datum/newscaster/feed_channel
 	var/admin_signature
 
+/datum/staff
```
ninjanomnom: this is basicaly just a duplicate of /datum/admins, make a shared parent

code/modules/admin/verbs/mentorhelp.dm
```diff
@@ -0,0 +1,561 @@
+/client/var/mentorhelptimerid = 0	//a timer id for returning the mhelp verb
+/client/var/datum/mentor_help/mentor_current_ticket	//the current ticket the (usually) not-mentor client is dealing with
+
+//
+//TICKET MANAGER
+//
+
+GLOBAL_DATUM_INIT(mhelp_tickets, /datum/mentor_help_tickets, new)
+
+/datum/mentor_help_tickets
```
ninjanomnom: After thinking about it a lot, tickets should all share a common parent rather than being a bunch of duplicated code like this. Do something like `/datum/tickets` holding most of the code and specific code can be in `/datum/tickets/admin` and `/datum/tickets/mentor`

Kmsxkuse: Problem is, the ticket stat counter fails to work with parenting admin and mentor. I tried my best to make it work but it doesn't work. If someone wants to try their hand at it, ping me on discord and i'll throw up what I have.

ninjanomnom:
Having 2 duplicated ticket systems in the code for mentors and non mentors is too much of a mess. If you're not willing to figure it out yourself someone else will just have to make a new pr at a later point.

Kmsxkuse:
Okay then.

Issue #1128, opened 2017-06-17
Bridge viewscreen doesn't work, doesn't open at all.


fizzle7:

ike709:
I've seen it work on the latest code.

fizzle7:
I keep shift clicking and changing view mode but it doesn't work, since the last update yesterday that is.

MASMC:
It didn't work for me either, as a borgie.

Dahlgaard:
There are multiple problems here.
The viewscreen itself has setup issues, and the viewscreen controller code is borked and was never intended for more than one viewscreen. Both viewscreen controllers end up linking to the viewscreen on the battlebridge. I've got a fix incoming for these things.

FalcoGer:
I can open the viewscreen window with right click + examine, but it doesn't work with shift clicking.

Dahlgaard:
@FalcoGer  I had experiencing the same as you report. The problems stems from the fact that the object has an "Initialize()" proc, which is supposed to call it's base proc, but fails to do so. Failing to do this ultimately means that a property on the object called "initialized" is never set, and when this isn't the case at a very low level in click.dm the click action is rejected early, and never sent through to other functions that are supposed to handle it. The PR i just posted fixes it, and takes care of some other issues i encountered while debugging this.

extramrdo:
Was this why the viewscreen was always "Almost falling apart!"

PR #1538, opened 2017-07-10
Fixes issue #1128, adds descs & makes controllers more flexible

Dahlgaard:
Fixes viewscreens #1128 and makes controllers more flexible

Added description to viewscreen & viewscreen_controller

Added base call to the Initialize() of viewscreen. 
    This is required for objects or they must set initialized to true themselves. 
    Failing to do this means you cannot click it as it will be rejected in click.dm. 
    This was causing problems referenced in issue #1128

Removed explicit Initialize() for viewscreen_controller. 
    This code was never set up to handle more then 1 viewscreen in the world, the current ship has 2. 
    The result was that all viewscreen_controllers were linked to the viewscreen on the backup battle bridge. 
    Possibly also part of items mentioned in issue #1128

Added helper proc get_area_viewscreens() which returns a list of viewscreens in the area.

Changed viewscreen_controller/attack_hand. 
    This will now popup a screen with a list of all viewscreens in the current area, allowing users to control them all from one viewscreen controller panel. 
    We can now have more viewscreens on the bridge if we want.

Changed viewscreen_controller/Topic. 
    Supports new actions in the ui. 
    Now has a usr.canUseTopic() check to ensure that non TK users cannot use the console from halfway across the ship.

### DIFF
--- a/code/game/objects/structures/viewscreen.dm
+++ b/code/game/objects/structures/viewscreen.dm
@@ -1,5 +1,6 @@
 /obj/structure/viewscreen
 	name = "bridge viewscreen"
+	desc = "This is a viewscreen display"
 	icon = 'icons/obj/96x96.dmi'
 	icon_state = "viewscreen"
 	density = 0
@@ -14,6 +15,7 @@
 
 
 /obj/structure/viewscreen/Initialize()
+	. = ..()
 	if(dir == 4 || dir == 8)
 		bound_width = 32
 		bound_height = 96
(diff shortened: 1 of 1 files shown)


Dahlgaard:
note sure if the CL part is necessary of filled out correctly feel free to wipe / edit it.

Dahlgaard:
CL part stripped after talks with ninja

code/game/objects/structures/viewscreen.dm
```diff
@@ -124,25 +126,28 @@
...
+	for(var/obj/structure/viewscreen/V in area_contents(A))
+		screens += V
+	return screens
 
 /obj/structure/viewscreen_controller/attack_hand(mob/user)
 	var/dat = "<B>Viewscreen Control Panel</B><HR>"
-	dat += "Viewing Mode: [linked.view_mode ? "Navigation" : "Tactical"]"
-	dat += "<BR><BR><A href=?src=\ref[src];switch=1>Switch View Modes</A>"
+
+	for(var/obj/structure/viewscreen/V in get_area_viewscreens())
```
ninjanomnom: You should not be re checking the area for viewscreen every time someone clicks the controller. Make a list on the controller holding all known screens instead and put a refresh button in the ui.

code/game/objects/structures/viewscreen.dm
```diff
@@ -124,25 +133,35 @@
 
 /obj/structure/viewscreen_controller
 	name = "viewscreen control panel"
+	desc = "This is a control panel for the viewscreens"
 	icon = 'icons/obj/airlock_machines.dmi'
 	icon_state = "airlock_control_standby"
 
 	density = 0
 	anchored = 1
+	var/list/linked_viewscreens = null
```
ninjanomnom: no need for `= null` here

code/game/objects/structures/viewscreen.dm
```diff
@@ -124,25 +133,35 @@
...
 
-/obj/structure/viewscreen_controller/Initialize()
-	for(var/obj/structure/viewscreen/V in world)
-		linked = V
+/obj/structure/viewscreen_controller/attack_hand(mob/user)
 
-	..()
+	// If this is being called for the first time, populate the list of screens
+	if (!linked_viewscreens)
+		linked_viewscreens = get_area_viewscreens()
```
ninjanomnom: do this in initialize

code/game/objects/structures/viewscreen.dm
```diff
@@ -151,8 +170,16 @@
 
 /obj/structure/viewscreen_controller/Topic(href,href_list)
 	..()
-	if(href_list["switch"])
-		linked.view_mode = !linked.view_mode
+	if(!usr.canUseTopic(src))
```
ninjanomnom: don't use `usr`

code/game/objects/structures/viewscreen.dm
```diff
@@ -151,8 +170,16 @@
...
+		return
+
+	if(href_list["refreshscreens"])
+		linked_viewscreens = get_area_viewscreens()
+
+	if(href_list["switchmode"])
+		var/obj/structure/viewscreen/V = locate(href_list["switchmode"])
```
ninjanomnom: instead of re locating the screen you should be passing the index of the screen in linked_viewscreens

ninjanomnom:
Let me know if/when you work on this again

PR #1611, opened 2017-07-17
Adds gay carrier ships that shoot loser simple mobs at the ship

Qustinnus:
:cl: 95% AuroranAI and 5% floyd / qustinnus
add: The syndicate found some old warehouses full of crappy outdated droids, they put them on big transport ships and are shooting them at NT ships now.
/:cl:

i never code mobs so dont yell too hard. they're using hivebot sprites right now. they dont do fuck all damage and are low hp as fuck so people can robust them easily. will need to test to see if I need to spawn more than 3 per hit.


code/datums/ftl_ships/ship_weapons.dm
```diff
@@ -1,8 +1,8 @@
 /datum/ship_attack
 	var/cname = "Ship Attack"
 
-	var/hull_damage = 0
-	var/shield_bust = 0
+	var/hull_damage = 0 //How much integrity damage an attack does
+	var/shield_bust = 0 //If it is blocked by shields
```
ninjanomnom: use TRUE/FALSE for booleans

code/datums/ship.dm
```diff
@@ -90,6 +94,12 @@ GLOBAL_VAR(next_ship_id)
 	for(var/i in init_components)
 		var/datum/component/component = SSship.cname2component(init_components[i])
 		var/datum/component/C = new component.type
+		if(istype(C, /datum/component/weapon/carrier_weapon) && (ship_boarders || num_boarders)) // we have a carrier weapon and something to change on it
+			var/datum/ship_attack/carrier_weapon/A = C.attack_data
+			if(ship_boarders)
+				A.boarding_mob = ship_boarders
+			if(num_boarders)
+				A.amount = num_boarders
```
ninjanomnom: object specific code in otherwise generic code is how you make me cry

code/modules/mob/living/simple_animal/hostile/droid.dm
```diff
@@ -0,0 +1,35 @@
...
+	ranged = 1
+	retreat_distance = 1
+	minimum_distance = 3
+	projectiletype = /obj/item/projectile/temp/droid
+	projectilesound = 'sound/weapons/laser.ogg'
+	gold_core_spawnable = 1
+	aggro_vision_range = 9
+	idle_vision_range = 2
+	turns_per_move = 5
+	atmos_requirements = list("min_oxy" = 0, "max_oxy" = 0, "min_tox" = 0, "max_tox" = 1, "min_co2" = 0, "max_co2" = 0, "min_n2" = 0, "max_n2" = 0)
```
AuroranAI: max tox should probably be zero, don't know why a robot that doesn't breathe would get hurt by breathing in plasma

code/datums/ship.dm
AuroranAI: hi, my name is idiot who didn't know that weapons had a reference to the ship they are bolted to

code/datums/ship.dm
```diff
@@ -12,8 +12,12 @@
 	var/recharge_rate = 0 // shield points per second
 
 	var/list/components = list()
+
 	var/salvage_map = "placeholder.dmm"
 
+	var/list/boarding_mobs = list(/mob/living/simple_animal/hostile/droid) // mob paths used in carrier_weapon, override with a specific mob path for something else
+	var/num_boarders = 5 // amount of npc boarders/shot to create
```
ninjanomnom: these should be in the weapon component

code/modules/mob/living/simple_animal/hostile/droid.dm
```diff
@@ -0,0 +1,35 @@
...
+	gold_core_spawnable = 1
+	aggro_vision_range = 9
+	idle_vision_range = 2
+	turns_per_move = 5
+	atmos_requirements = list("min_oxy" = 0, "max_oxy" = 0, "min_tox" = 0, "max_tox" = 0, "min_co2" = 0, "max_co2" = 0, "min_n2" = 0, "max_n2" = 0)
+	damage_coeff = list(BRUTE = 0.75, BURN = 1, TOX = 0, CLONE = 0, STAMINA = 0, OXY = 0)
+	a_intent = INTENT_HARM
+	harm_intent_damage = 5
+	speak_emote = list("beeps")
+	faction = list("syndicate")
```
ninjanomnom: probably want to set this based on ship faction

AuroranAI: Dealt with, faction is now set while the weapon is fired.

code/datums/ftl_ships/ship_weapons.dm
```diff
@@ -146,6 +146,33 @@
...
+		var/mob/to_spawn = new path(epicenter)
+		to_spawn.faction = list(ship.faction)
+
+/datum/ship_attack/carrier_weapon/oneTime
+	var/fired = FALSE
+
+/datum/ship_attack/carrier_weapon/oneTime/damage_effects(turf/epicenter)
+	if(!fired)
+		..()
+		fired = TRUE
```
AuroranAI: looking back at this code, i'm probably going to add the onetime variable and another one to the parent instead, so you can have the code both look for it firing once and then have a specific variable to have it go through the list and spawn all of the mobs specifically instead of still doing it random like it is here. I'd like to do this in another PR though, since this one is essentially fully functional.

code/datums/ship.dm
```diff
@@ -12,8 +12,12 @@
 	var/recharge_rate = 0 // shield points per second
 
 	var/list/components = list()
+
 	var/salvage_map = "placeholder.dmm"
 
+	var/list/boarding_mobs = list() // mob paths used in carrier_weapon, override with a specific mob path for something else
+	var/num_boarders = 5 // amount of npc boarders/shot to create
```
ninjanomnom: don't forget to remove this now that it's been moved to the weapon component

code/datums/ship.dm
```diff
@@ -97,6 +101,8 @@ GLOBAL_VAR(next_ship_id)
 		C.x_loc = text2num(coords[1])
 		C.y_loc = text2num(coords[2])
 		C.ship = src
+		if(C.attack_data)
+			C.attack_data.ship = src
```
ninjanomnom: move this into the component. This is component specific code in a generic proc. Pass src in the new proc honestly.

AuroranAI: Moving this into component would require adding a callback, which you told me not to do. Passing the src in New would lead to race conditions since the ship is set in this code. What I could do is add the component to the attack data instead and just daisy chain up the list to get the faction reference?

ike709:
It's a transport not a carrier reeeeeee

code/datums/ftl_ships/ship_weapons.dm
```diff
@@ -146,6 +146,33 @@
...
+	var/list/boarding_mobs = list(/mob/living/simple_animal/hostile/droid)
+	var/amount = 5
+
+/datum/ship_attack/carrier_weapon/damage_effects(turf/epicenter)
+
+	playsound(epicenter, 'sound/ftl/shipweapons/carrier_hit.ogg', 100, 1)
+	for(var/I = 1 to amount)
+		var/path = pick(boarding_mobs)
+		var/mob/to_spawn = new path(epicenter)
+		to_spawn.faction = list(our_ship_component.ship.faction)
```
Super3222: `our_ship_component.ship.faction` is not an actual variable it seems. I don't know how this is suppose to work, but is line 11 (`/var/datum/component/our_ship_component`) actually suppose to be `datum/ship_component/our_ship_component`?

Qustinnus: Yes it is, thanks. the reason this is happening by the way is because I changed all components to ship_component and this had conflicts and i lazily resolved them :^)

PR #1661, opened 2017-07-22
Changes latejoin to allow for multiple back up areas.

AuroranAI:
Before the rebase, even if all the chairs were destroyed, you could respawn in both hallways. This adds that functionality back by also allowing mappers to designate a custom area/areas to use as their back up latejoin area. It defaults to the arrivals hallway if you don't place down a landmark. 

:cl: AuroranAI
add: In case of all the chairs getting destroyed, you can now respawn in both the arrivals and the exit hallway of the Aetherwhisp.
/:cl:


code/controllers/subsystem/job.dm
```diff
@@ -529,7 +530,10 @@ SUBSYSTEM_DEF(job)
 		SendToAtom(M, pick(latejoin_trackers), buckle)
 	else
 		//bad mojo
-		var/area/shuttle/ftl/hallway/secondary/entry/A = locate() in GLOB.sortedAreas
+		if(!latejoin_area_trackers.len)
+			var/area/shuttle/ftl/hallway/secondary/entry/A = locate() in GLOB.sortedAreas
+			latejoin_area_trackers += A  // default to the arrival hallway if a mapper didn't set up a landmark
+		var/area/A = locate(pick(latejoin_area_trackers)) in GLOB.sortedAreas
```
ninjanomnom: use an area initialize instead of a landmark

PR #1669, opened 2017-07-22
Fixes shuttle leaving behind thrusters

Qustinnus:
thanks dad


code/modules/shuttle/on_move.dm
```diff
@@ -189,6 +189,14 @@ All ShuttleMove procs go here
 	var/datum/camerachunk/chunk = GLOB.cameranet.getCameraChunk(x, y, z)
 	chunk.hasChanged(TRUE)
 
+/obj/structure/shuttle/beforeShuttleMove()
```
ninjanomnom: this goes in the structure section, not the machinery section of the file

Qustinnus: these are bimmers thrusters. the ones on the mining shuttle

PR #1680, opened 2017-07-22
Fix the hub not working

monster860:
So I pulled out my small BYOND environment to test it out and it turns out it actually works


Qustinnus:
hostchat leak!!! ban he!!!!

MrStonedOne:
Now push it upstream to /tg/!

PR #1687, opened 2017-07-23
[DNM] [WIP] Random events that make the game less random in an ironic way.

Qustinnus:
#This PR removes a fuckton, including current ship spawning and star economy, my reasons for that are simple; it's not being used, it over complicates an already un-fun system, and it's just not really helping us towards our goals.

I'm keeping fleet movement but it should be refactored removing things such as calling for help. So we can use it for things such as actual roaming fleets etc, but instead make these random instead of 

Combat events: Allow you to pick fight avoid or bribe, only fight is implemented right now. Engaging in combat is done automatically if nothing is picked.

Ruin events: Spawn a new docking point with a ruin map

Quest events: Will be done in another PR

Passive events: Will have X amounts of choice with Y effects for each button. Could also just be 1 button which is "accept". If no button is pressed it should just go with the default option after a while.

To do:
- [x] Remove current implementation of ship 

- [x]  Lists of events gets made based on what factions it can spawn 

- [x] Every planet gets an event at round start

- [x] On arriving at a new planet that planets event gets activated

- [x] Combat events functionality

- [x] Ruin events 

- [ ] Signal scanner machine

- [ ] Passive events

- [ ] UI (Shows all visible events on the console, this allows events like pirate combat events to be invisible till encountered, but you can scan for things like ruins beforehand) 50% done

Things I didn't mean to do but did anyways

- [x] Learn about docking ports and mapping in FTL

- [x] Learn how to use TGUI

- [x] Fix the FTL nav console to make it user friendly (after 1 year)

- [x] Make the jump range configurable on the ftl drive


Ecolli:
We want to remove the huge "fleet battles" and move towards more epic 1v1 type combat. Our ship should NOT be able to spam fire a fleet of ships into oblivion. This will be a step towards making combat engagements more impactful and daunting.

Qustinnus:
@ninjanomnom please look at the UI code I just don't know how to do it properly.

Qustinnus:
@ninjanomnom can i have a review of what i have so far

code/controllers/subsystem/starmap.dm
```diff
@@ -135,11 +160,16 @@ SUBSYSTEM_DEF(starmap)
 			current_planet = current_system.planets[1]
 		else if(in_transit_planet)
 			current_planet = to_planet
+			current_planet.event.init_event()
```
ninjanomnom: Unless you don't want events when jumping systems, this shouldn't be in this conditional block

code/controllers/subsystem/starmap.dm
```diff
@@ -135,11 +160,16 @@ SUBSYSTEM_DEF(starmap)
...
+
 
 		var/obj/docking_port/stationary/dest = current_planet.main_dock
 		ftl.mode = SHUTTLE_CALL
 		ftl.destination = dest
 
+		current_system.visited = 1
```
ninjanomnom: This is already on line 159

code/controllers/subsystem/starmap.dm
```diff
@@ -149,6 +179,22 @@ SUBSYSTEM_DEF(starmap)
 
 		from_time = 0
 		to_time = 0
+
+		in_transit_planet = 0
```
ninjanomnom: this is already done on line 202 and 203. If you need to move it up here, remove the old one.

code/controllers/subsystem/starmap.dm
```diff
@@ -149,6 +179,22 @@ SUBSYSTEM_DEF(starmap)
...
+
+		sleep(1)
```
ninjanomnom: I rewrote this proc to remove the sleep and spawn, now you add it back in? ;v;

code/controllers/subsystem/starmap.dm
```diff
@@ -149,6 +179,22 @@ SUBSYSTEM_DEF(starmap)
...
+
+		for(var/area/shuttle/ftl/F in world)
+			F << 'sound/effects/hyperspace_end.ogg'
+		toggle_ambience(0)
```
ninjanomnom: This is already done on line 173, and with better performance

code/controllers/subsystem/starmap.dm
```diff
@@ -149,6 +179,22 @@ SUBSYSTEM_DEF(starmap)
...
+
+		spawn(50)
+			ftl_sound('sound/ai/ftl_success.ogg')
```
ninjanomnom: This is already done on line 178. Just in a way to not have a spawn

code/controllers/subsystem/starmap.dm
```diff
@@ -149,6 +179,22 @@ SUBSYSTEM_DEF(starmap)
...
+
+		if(current_planet.visited != TRUE)
+			current_planet.event.activate_event()
+			current_planet.visited = TRUE
```
ninjanomnom: Put visited under the conditional block of line 161. You don't need to check if it's not true, just set it to TRUE and be done with it.

ninjanomnom: And if activate event needs to be done later, use a timer

code/modules/FTL/consoles/ftl_navigations.dm
```diff
@@ -0,0 +1,310 @@
...
+	var/icon/planet_icon
+	var/selecting_planet = 0
+	var/do_send
+	var/icon_view_counter = 0
+	var/secondary = FALSE //For secondary Battle Bridge computers
+	var/general_quarters = FALSE //Secondary computers only work during General Quarters
+
+
+/obj/machinery/computer/ftl_navigation/New()
+	..()
```
ninjanomnom: `. = ..()`

code/modules/FTL/consoles/ftl_navigations.dm
```diff
@@ -0,0 +1,310 @@
...
+	if(secondary)
+		name = "secondary navigation console"
+		desc = "This is a backup navigation console. It will only work during General Quarters."
+		icon = 'icons/obj/computerold.dmi' //old nasty sprite for a secondary computer
+		icon_keyboard = "teleport_key" //so it fits the old sprite
+		icon_screen = "navigation" //so it has a screen
+
+	SSstarmap.ftl_consoles += src
+	spawn(5)
```
ninjanomnom: stop

code/modules/FTL/consoles/ftl_navigations.dm
```diff
@@ -0,0 +1,310 @@
...
+		selected_system = SSstarmap.current_system
+		screen = 1
+
+/obj/machinery/computer/ftl_navigation/Destroy()
+	SSstarmap.ftl_consoles -= src
+	.=..()
```
ninjanomnom: Why must you format it different

code/modules/FTL/consoles/ftl_navigations.dm
```diff
@@ -0,0 +1,310 @@
...
+
+/obj/machinery/computer/ftl_navigation/ui_interact(mob/user, ui_key = "main", datum/tgui/ui = null, force_open = 0, datum/tgui/master_ui = null, datum/ui_state/state = GLOB.default_state)
+	if(secondary && !general_quarters)
+		user << "This console is locked. Backup consoles only work during General Quarters."
```
ninjanomnom: to_chat

code/controllers/subsystem/starmap.dm
```diff
@@ -33,21 +33,45 @@ SUBSYSTEM_DEF(starmap)
...
+	for(var/etype in subtypesof(/datum/ftl_event))
+		var/datum/ftl_event/event = new etype()
+		if(!event.name)
+			continue
+		switch(event.faction)
+			if(FTL_NEUTRAL, FTL_PIRATE) //Pirates are considered a neutral event
+				neutral_events[etype] = event.rarity
+				solgov_events[etype] = event.rarity
+				syndicate_events[etype] = event.rarity
+				nanotrasen_events[etype] = event.rarity
```
Qustinnus: neutral events can spawn in their sectors.

code/controllers/subsystem/starmap.dm
Qustinnus: i think it was in when I started on this

Qustinnus: i wouldnt add a sleep(1) because i have no clue why itd be there

code/modules/FTL/consoles/ftl_navigations.dm
Qustinnus: I DIDNT DO THIS

code/modules/FTL/consoles/ftl_navigations.dm
Qustinnus: I DIDNT DO THIS XD

code/controllers/subsystem/starmap.dm
```diff
@@ -33,21 +33,45 @@ SUBSYSTEM_DEF(starmap)
...
+			if(FTL_SOLGOV)
+				solgov_events[etype] = event.rarity
+			if(FTL_SYNDICATE)
+				syndicate_events[etype] = event.rarity
+			if(FTL_NANOTRASEN)
+				nanotrasen_events[etype] = event.rarity
+			else
+				all_events[etype] = event.rarity
```
Qustinnus: ok

code/controllers/subsystem/starmap.dm
Qustinnus: pickweight wont accept multiple lists i think

ninjanomnom: `pickweight(neutral_events | solgov_events)`

code/controllers/subsystem/starmap.dm
```diff
@@ -149,6 +175,12 @@ SUBSYSTEM_DEF(starmap)
 
 		from_time = 0
 		to_time = 0
+
+		if(current_planet.visited != TRUE)
+			current_planet.event.activate_event()
```
ninjanomnom: Currently activate_event is called immediately after init_event with no delay. I'm not sure if you intend that.

Regardless, move this conditional to above with the other conditionals. Also the way it currently is, if a planet is returned to it will call init_event again on the event even if activate_event isn't.

code/controllers/subsystem/starmap.dm
```diff
@@ -450,7 +420,7 @@ SUBSYSTEM_DEF(starmap)
 					if(!prob(f_list[ship_to_spawn]))
 						ship_to_spawn = null
 
-				SSship.create_ship(ship_to_spawn,faction.cname,capital_system)
+				SSship.create_ship(ship_to_spawn,faction.cname,capital_system) */
```
ninjanomnom: What's the point in completely removing some things and commenting others? Just remove it all, we have a code history for a reason.

code/controllers/subsystem/starmap.dm
```diff
@@ -577,3 +529,20 @@ SUBSYSTEM_DEF(starmap)
...
+			event = pickweight(solgov_events)
+		if(FTL_SYNDICATE)
+			event = pickweight(syndicate_events)
+		if(FTL_NANOTRASEN)
+			event = pickweight(nanotrasen_events)
+		else
+			event = pickweight(all_events)
+
+	var/datum/ftl_event/new_event = new event
+	return new_event
```
ninjanomnom: I'd recomend you make this just return the type and not an instanced event in case you want to add New args in the future.

code/datums/stars.dm
```diff
@@ -115,6 +116,8 @@
 	disp_level = index
 	disp_angle = rand(0, 360)
 	map_names = list()
+	event = SSstarmap.get_new_event(parent_system.alignment)
+	event.our_planet = src
```
ninjanomnom: Like this for example could be a New arg.

code/modules/FTL/events/event_combat.dm
```diff
@@ -0,0 +1,36 @@
...
+	description = "A pirate ship is spotted in orbit of the planet"
+	rarity = COMMON_EVENT
+	faction = FTL_PIRATE
+	ships_to_spawn = list(/datum/starship/clanker=1)
+
+/datum/ftl_event/combat/random
+	name = "Pirate ships"
+	description = "An assortment of pirate ships is spotted in orbit of the planet, It seems like they stole these ships somewhere."
+	faction = FTL_PIRATE
+	rarity = 	UNCOMMON_EVENT
```
ninjanomnom: nice formatting

tgui/src/interfaces/ftl_navigation.ract
```diff
@@ -1,12 +1,16 @@
 {{#if data.screen == 0}}
+	<ui-button icon='info-circle' action='shipinf' state='{{data.can_jump ? null : "selected"}}' >Ship Information</ui-button>
```
ninjanomnom: I'm not going to review ract files, this is a mess in the first place and if you've made changes that work then it's fine

code/controllers/subsystem/starmap.dm
Qustinnus: me being salty, I intended on removing it all after I realized the same thing.

code/modules/FTL/events/event_combat.dm
Qustinnus: this wizardry happend behind my back.

tgui/src/interfaces/ftl_navigation.ract
Qustinnus: it's the only part thats WIP but its basic functionality works.

code/modules/FTL/events/event_actions/combat_actions.dm
```diff
@@ -0,0 +1,35 @@
...
+		flavortext = "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA WE GOT SPOTTED"
+		our_event.spawn_ships()
+
+
+/datum/ftl_event_action/bribe
+	buttonname = "Attempt to bribe them"
+	var/bribe_cost = 1000
+
+/datum/ftl_event_action/bribe/activate() //Called when you press the button.
+	if(prob(75 && SSshuttle.points > 1000))
```
F-OS: Shouldn't this be prob(75) && SSshuttle.points > 1000?

Qustinnus: yea it should but this was mainly psuedocode and it's an old commit

crossedfall:
Closing in favor of 2.0

PR #1707, opened 2017-07-25
Smooths out TG Movement

Xantholne:
Smooths out movement making mob movement look and feel a lot smoother and less like a game of chess.

:cl: Xantholne
fix: Movement is now less choppy and a lot smoother
/:cl:

### DIFF
--- a/code/game/atoms_movable.dm
+++ b/code/game/atoms_movable.dm
@@ -30,7 +30,6 @@
 	var/moving_diagonally = 0 //0: not doing a diagonal move. 1 and 2: doing the first/second step of the diagonal move
 	var/list/client_mobs_in_contents // This contains all the client mobs within this container
 	var/list/acted_explosions	//for explosion dodging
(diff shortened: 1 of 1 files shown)


Qustinnus:
@ninjanomnom this shouldnt cause any issues, right?

ninjanomnom:
I don't know off hand what this would do and I can't investigate it or test it.

Xantholne:
It makes it so movement isn't like a game of chess and more smooth. Currently running it right now on my private while checking out RND

ninjanomnom:
I'm talking about from a code side of things I don't know how that var is used. Plenty of things exist in the code which can be toggled to improve the feel or look of the game but are set how they are because of technical reasons.

Xantholne:
From what we know code wise, it should just be controling how many pixels you glide over when you move.

ninjanomnom:
After testing it myself the movement just gets more jerky

PR #1799, opened 2017-08-02
Mirror PR: Made some object lists less terrible to read

AuroranAI:
Also made me have no will to live because FUCK
closes #1633
Note: This is not ~entirely~ done. Map_template.dm.rej is not changed, because we don't actually have the changes in our code. If you know where this is in the rest of the code/if it actually matters ninja, poke me on discord.


code/modules/mapping/map_template.dm.rej
```diff
@@ -0,0 +1,14 @@
+diff a/code/modules/mapping/map_template.dm b/code/modules/mapping/map_template.dm	(rejected hunks)
+@@ -32,10 +32,10 @@
```
ninjanomnom: You missed a reject

PR #1815, opened 2017-08-02
Mirror PR: Augment Styles/Changes

AuroranAI:
Closes #1628

🆑 Dannno/Supermichael777/InsaneHyena
add: You can now pick from a few different styles when augmenting someone with robot parts by putting them in the augment manipulator. Alt+click to take parts out.
/🆑

image

Full augment sets on the left, partial mix/match on the right.

This PR aims to make augmenting slightly more interesting than just turning your body grey, while also begin reminiscent of the cyborg parts that are actually used in the process.

Using the augment manipulator now found in robotics, one can insert a part and custom fit it to one of three different styles (security, engineering, and standard). Mix and match if you so choose, or make a full set for maximum style.


ike709:
An augment manipulator needs to be mapped in at robotics if you haven't already.

Qustinnus:
that's fucking sexy

AuroranAI:
Going to let Aer map it in, this has iit on the code side.

PR #1842, opened 2017-08-04
Fixes stechkins for security [hotfix]

Cdey78:
makes security actually spawn with stechkins, not invisible "Projectile Gun"s and makes them spawn with the correct ammo type

Does this really need a changelog?


Cdey78:
ah, fuck
I goofed

PR #1857, opened 2017-08-05
Fixes ships being unable to shoot at players, and players being unable to shoot back

AuroranAI:
Ship components now generates it's datum list properly, and the tactical console now properly gets the component list.
Also did a wee bit of tidying up while i was here, and turned some typesof into subtypesof.
:cl: AuroranAI
fix: The Space Christmas Truce of 2914 is now over. Get back to work.
/:cl:

### DIFF
--- a/code/controllers/subsystem/ships.dm
+++ b/code/controllers/subsystem/ships.dm
@@ -28,17 +28,17 @@ SUBSYSTEM_DEF(ship)
 
 
 /datum/controller/subsystem/ship/proc/init_datums()
-	var/list/factions = typesof(/datum/star_faction) - /datum/star_faction
+	var/list/factions = subtypesof(/datum/star_faction)
 
 	for(var/i in factions)
 		star_factions += new i
 
-	var/list/ship_components = typesof(/datum/ship_component) - /datum/ship_component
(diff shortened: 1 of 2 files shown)


code/controllers/subsystem/ships.dm
```diff
@@ -28,17 +28,17 @@ SUBSYSTEM_DEF(ship)
 
 
 /datum/controller/subsystem/ship/proc/init_datums()
-	var/list/factions = typesof(/datum/star_faction) - /datum/star_faction
+	var/list/factions = subtypesof(/datum/star_faction)
```
Qustinnus: better coding standards +1

code/game/machinery/computer/ftl_weapons.dm
```diff
@@ -127,7 +127,7 @@
 		if(target)
 			data["target"] = target.name
 			var/list/ship_components_list = list()
-			data["ship_components"] = ship_components_list
+			data["components"] = ship_components_list
```
Qustinnus: good enough for now we might want to change the .ract at some point but I can put it on my to-do-list for further combat revamp updates.

PR #1859, opened 2017-08-05
Admin smite messages

Maciekkub:
:cl: Smitting
Added an abbility for admins to choose what message they will display on the smitten player
Added the default message for every smite type when you don't input a message
/:cl:

[why]:  A minor thingy to make the game more clear. Now instead of just taking those 75 brain damage for example, you will also get a message, either chosen by the admin who smitted you, or the default one!

### DIFF
--- a/code/modules/admin/verbs/randomverbs.dm
+++ b/code/modules/admin/verbs/randomverbs.dm
@@ -1188,16 +1188,24 @@ GLOBAL_LIST_EMPTY(custom_outfits) //Admin created outfits
 			T.Beam(target, icon_state="lightning[rand(1,12)]", time = 5)
 			target.adjustFireLoss(75)
 			target.electrocution_animation(40)
-			to_chat(target, "<span class='userdanger'>The gods have punished you for your sins!</span>")
+			if(!message)
+				message = "The gods have punished you for your sins!"
+			to_chat(target, "<span class='userdanger'>[message]</span>")
 		if(ADMIN_PUNISHMENT_BRAINDAMAGE)
-					if(!message)
+			if(!message)
 							message = "You feel dumber"
-					target.adjustBrainLoss(75)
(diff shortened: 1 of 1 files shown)


code/modules/admin/verbs/randomverbs.dm
```diff
@@ -1176,7 +1176,7 @@ GLOBAL_LIST_EMPTY(custom_outfits) //Admin created outfits
 		return
 
 	var/list/punishment_list = list(ADMIN_PUNISHMENT_LIGHTNING, ADMIN_PUNISHMENT_BRAINDAMAGE, ADMIN_PUNISHMENT_GIB, ADMIN_PUNISHMENT_BSA)
-
+	var/message = input("Choose the message they will receive") as text
```
Qustinnus: I was going to say put var/message under var/punishment, but you should actually put var/message under if(QDELETED(target) || !punishment) so it only asks what message if the user has entered a punishment type.

code/modules/admin/verbs/randomverbs.dm
```diff
@@ -1188,13 +1188,24 @@ GLOBAL_LIST_EMPTY(custom_outfits) //Admin created outfits
...
 		if(ADMIN_PUNISHMENT_GIB)
+			if(!message)
+				message = "The gods have punished you for your sins!"
 			target.gib(FALSE)
+			to_chat(target, "<span class='userdanger'>[message]</span>")
 		if(ADMIN_PUNISHMENT_BSA)
+			if(!message)
+				message = "It seems you have angered Centcom a bit too much"
 			bluespace_artillery(target)
+			to_chat(target,"<span class='userdange'>[message]</span>")
```
Qustinnus: userdange

Qustinnus: maciekkub.adjustBrainLoss(75)

PR #1888, opened 2017-08-08
mirror pr | Ore satchels no longer have infinite capacity

Qustinnus:
Original PR: tgstation/tgstation#29401

🆑 Xhuis
fix: Mining satchels no longer hold infinite amounts of ore.
/🆑

Fixes #29392.

I also ran into an issue while working on this that would flood the chat log with messages saying that there wasn't enough space. To get around this I added a var called spam_protection to mining satchels. Trying to pick up a piece of ore will send an error message and then set spam_protection = TRUE, then to FALSE a decisecond later - while it's TRUE error messages of that kind can't appear.

closes #1657


code/modules/mining/ores_coins.dm.rej
```diff
@@ -0,0 +1,25 @@
+diff a/code/modules/mining/ores_coins.dm b/code/modules/mining/ores_coins.dm	(rejected hunks)
+@@ -23,6 +23,7 @@
```
ninjanomnom: you missed a reject

PR #1891, opened 2017-08-08
pr mirror | Changes clickcatcher to dynamically scale based on view range rather than making 1 tile for each tile viewed

Qustinnus:
Original PR: tgstation/tgstation#29370

Scale()s the icon when range is below 16, if it's above 16, scales the icon to 16 and transforms the rest of the way.
fixes #29342

closes #1686


ninjanomnom:
fix dat travis

Qustinnus:
@ninjanomnom fixed

PR #1982, opened 2017-08-16
Mirror Fix: Revert the hive bots event

Funce:
:cl: Funce
del: The Hivebots no longer invade
/:cl:

closes #1733 
If that doesn't close it then it needs to be closed after this one is merged.
I left the assets in. I can remove them also if required.


Qustinnus:
Why exactly does this need to be removed?

Qustinnus:
Oh nevermind, fix the mirror PR instead, its easier, prevents later conflicts, and  tbf is a bit cleaner.


message me if u want the teachings

Funce:
Uh yeha, Do not merge whilst I attempt to do just that

PR #1986, opened 2017-08-16
Adds ability for players to lower their head into arms.

The-clown-known-as-Pepe:
Adds dabbing.

http://i.imgur.com/dZsTGIv.png

http://i.imgur.com/b8stKJp.png

### DIFF
--- a/code/modules/mob/living/emote.dm
+++ b/code/modules/mob/living/emote.dm
@@ -18,12 +18,6 @@
 /datum/emote/living
 	mob_type_allowed_typecache = list(/mob/living)
 	mob_type_blacklist_typecache = list(/mob/living/simple_animal/slime, /mob/living/brain)
-
-/datum/emote/living/dab
-	key = "dab"
-	key_third_person = "dabs"
-	message = "dabs."
-	L.gib()
 	
 /datum/emote/living/blush
 	key = "blush"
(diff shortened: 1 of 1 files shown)


nichlas0010:
image image

The-clown-known-as-Pepe:
@Qustinnus

ninjanomnom:
You can do this already with custom emotes

The-clown-known-as-Pepe:
that is a argument to delete all existing emotes and not change anything ever

ninjanomnom:
Emotes *do* exist that add additional sound or animation. But yes, emotes that don't add anything besides default behavior are pointless

The-clown-known-as-Pepe:
Added sound, animation and effect to dabbing as per "requested".

ninjanomnom:
Self gibbing emotes isn't the kind of thing we do here.

The-clown-known-as-Pepe:
oh then i'll revert it to the original state

ninjanomnom:
As I said in coderbus, you'll need a good reason for this sort of thing to get merged, or convince ecolli or floyd to support it. Code-wise just the base emote has no reason to exist because custom emotes can already be done

Qustinnus:
the chances of me supporting this are as high as your chances of becoming an admin.


read: not very high.

ninjanomnom:
Yeah this isn't happening

PR #2039, opened 2017-08-20
[MIRROR FIX] Moved emagged var to obj/item parent.

Funce:
Closes #2031 

No changelog needed. (Only an invisible refactor)

Its a mirror fix

Also, there is a missing file, probably pending another mirror PR.
`code/game/objects/items/circuitboards/computer_circuitboards.dm`

Well it seems the debugging found these. Plus one more.


Funce:
Oh dear ah, I think something happened

PR #2043, opened 2017-08-20
Removes bridge doors

stobarico:
Adds normal bridge doors (Command ones)


ninjanomnom:
Update this and I'll merge it

stobarico:
I will close this because it has conflics, Aerwhisp is updating his map, and it isn't mapmerged properly

PR #2061, opened 2017-08-23
Fix cameranet bugs

NIXC:
Now instead of trying to update the chunks when the ship moves, we instead throw our hands up in the air and toss the chunks out too. Then we rebuild them as they are needed. If there are no AI eyes the chunks should never be rebuilt (which should reduce overhead).

:cl: NINXE
fix: AI camera bug finally dead
/:cl:

I've tried to test this as best as I can, but having other people try to break it would be nice. I'm still a little sketchy on how shuttles work.


NIXC:
Shouldnt. If the camera is outside the shuttle, it wont get removed from the cameranet (no beforeshuttlemove()), and thus wont be added back in. The only thing it will lose is its chunk which if an AI eye passes through will get rebuilt. I don't think camera consoles use chunks anyway, but I may be mistaken, Ill have to check.

Thats probably all moot though, are we ever going to have cameras not on the ship that actually work? As it stands if stations have cameras the AI cant use them anyway.

ninjanomnom:
It's not a matter of whether we'll ever use it, it's a bad fix that breaks something else. Also as I said in coderbus: This is a bad fix regardless because you shouldn't have to delete all the chunks for cameras to work.

PR #2124, opened 2017-08-28
Ports Special Intents from Yog

Super3222:
[Changelogs]:

:cl: Super3222
fix: Fixes that annoying say bug where () shows up every time you talk.
/:cl:


Super3222:
image of this working:

Funce:
Title is a little Sketchy

code/game/say.dm
```diff
@@ -71,7 +71,7 @@ GLOBAL_LIST_INIT(freqtospan, list(
 	return ""
 
 /atom/movable/proc/compose_job(atom/movable/speaker, message_langs, raw_message, radio_freq)
-	return "[" (" + speaker.GetJob() + ")"]"
+	return "[ speaker.GetJob() ? " (" +  speaker.GetJob() + ")" : ""]"
```
Funce: I would suggest instead of calling the GetJob proc twice, to make it into a var to reference twice.

Super3222: Sorry I have a single-line mentality

Super3222:
I feel like this PR is broken, so I'm moving it to https://github.com/FTL13/FTL13/pull/2217.

PR #2193, opened 2017-09-03
Fixes ruins spawning inside the ship area

ninjanomnom:
:cl: ninjanomnom
fix: Ruins should no longer spawn in the ship area
/:cl:

Because of how map gen works we have to use new here unfortunately unless we want to do quite a bit larger of a rewrite.

### DIFF
--- a/code/modules/shuttle/shuttle.dm
+++ b/code/modules/shuttle/shuttle.dm
@@ -208,9 +208,12 @@
 
 	var/planet_dock = FALSE //var to help with escape pod landings
 
+/obj/docking_port/stationary/New()
+	. = ..()
+	SSshuttle.stationary += src //This has to be here to pre-empt ruin spawning
+
 /obj/docking_port/stationary/Initialize()
 	. = ..()
-	SSshuttle.stationary += src
 	if(!id)
 		id = "[SSshuttle.stationary.len]"
 	if(name == "dock")


code/modules/shuttle/shuttle.dm
```diff
@@ -208,9 +208,12 @@
 
 	var/planet_dock = FALSE //var to help with escape pod landings
 
+/obj/docking_port/stationary/New()
+	. = ..()
+	SSshuttle.stationary += src //This has to be here to pre-empt ruin spawning
+
```
monster860: This will not work if this is on a compiled-in map because it will reference a non-existent SSshuttle

ninjanomnom: This is how it originally was back when we *did* use compiled maps, so no I'm pretty sure it should still work.

Testing it myself there are no runtimes when the map is compiled in like it would if SSshuttle were null.

ncaaseawolf: does not
work

ncaaseawolf:
Does not work.
A ruin spawned right inside of the ship at round start.

ninjanomnom:
It's not actualy testmerged atm, the server is lying about it for some reason

ninjanomnom:
Testmerged twice and no issues

PR #2326, opened 2017-09-12
MIRROR FIX: Makes GetComponent() faster using typecache magic

AuroranAI:
I dream of a day our children are not plagued by thousands of mirror prs
Closes #1793 

"Added /proc/type2parent which takes a type (string, or type) and returns it's parent type (initial(x.parent_type) doesn't work)

Instead of being a raw list of components, /datum/var/datum_components is now associated list keyed by type with some bells and whistles to make GetComponent() calls faster

e.g.

/datum/component/a
    ...

/datum/component/b
    ...

/datum/component/b/c
    ...

var/list/datum_components = list(
    /datum/component/a = <instance of a>,
    /datum/component/b = list(<instance of b>, <instance of b>, <instance of c>),
    /datum/component/b/c = list(<instance of c>, <instance of c>)
)"


ninjanomnom:
You'll need more maintainers for that

PR #1908, opened 2017-08-09
[ready4freddy] Cargo overhaul

AuroranAI:
It's about time I put this up so other people can look at it and critique it. The google docs has all of my current working changes. A lot of numbers in this are subject to change; most were in for development purposes, and don't reflect when an actual balance pass happens. 

https://docs.google.com/document/d/1Ajiyv2XDPYlhq8eH9TiYVOTL8LL8rkYK0D9JZPuBkNE/edit#

:cl: AuroranAI
add: Stations now have a module that determines what objects they'll have on sale. It will also determine prices of both selling and buying. 
tweak: Stations will have more types of supplies on sale, and will have more of them. 

/:cl:

PR #2382, opened 2017-09-14
Adds a new station styled after a chopshop

TMTIME:
:cl: TMTIME
add: A new station, intended to be used in #1908 
tweak: Allows the new station to appear normally
/:cl:

[why]: The cargo overhaul (#1908) means that stations can have varying stock and prices. Giving each station it's own map is neat for two reasons:
1 - Helps players figure out the kind of items the station would sell or buy at decent prices
b - Shops like the clothing store don't really need a xenobiology.
iii - Adds more variation to the stations

This station is built to fill the chopshop module, and is themed after the buy and sell keywords. As far as powergaming prevention, there is so giftshop onboard.

Also I was told to remove the protective dock so that happened.

Here is a screenshot of the station *before* I removed the protective dock
https://cdn.discordapp.com/attachments/350840634925645827/357645342956584960/station_chopshop.png

~~i probably did something wrong submitting this r i p~~


AuroranAI:
I mean this should probably wait until after the cargo overhaul is actually merged, and the way you did this currently makes it so pizza places can be chop shops, but i can probably handle that in a separate pr

TMTIME:
I'm okay waiting. Also I only have it set to load as any station so it can be tested to see if it breaks/for feedback from players during a round.

ninjanomnom:
As @AuroranAI said this should wait on the other pr to be merged

PR #2486, opened 2017-09-20
Fix podlocks appearing under poddoors that you don't want to have podlocks under

monster860:
Ecolli was getting really annoyed by this


Ecolli:
Thanks for looking into this Monster. It'll be worth it!

PR #2518, opened 2017-09-23
Ecolli's Scarab

ninjanomnom:
:cl: Ecolli
add: A new ship: The Scarab
tweak: Map rotation has been enabled
/:cl:

I'll let @Ecolli elaborate if he feels the need.


TMTIME:
bug?: everyone spawned in the bar

TMTIME:
Told to put all issues here so here we go:

**General miner access issues**
Airlocks that are locked to miners that they might need access to:
- External airlocks connecting to the rear of the mining/FOB shuttle
- Cockpit airlock
- ORM (Currently the ore drop point is in cargo. If the pop is high, then miners do not start with cargo access)

**Escape shuttle is behind external access airlocks**
tldr external access is a command access only, so only command will be able to open the doors.

PR #2519, opened 2017-09-23
Higher boarding droid projectile damage

ninjanomnom:
:cl: ninjanomnom
balance: Boarding droids do more ranged damage (+66%)
/:cl:


ninjanomnom:
Let's just merge this and see if we get people dying, if not try again

PR #2529, opened 2017-09-24
removes the empty level spawning in SSmapping

ninjanomnom:
reduces some memory usage, testmerging first to make sure it breaks nothing


Qustinnus:
content freeze pls

PR #2545, opened 2017-09-25
FOB Shuttle

ike709:
🆑 ike709, not_ike709, Floyd
add: The mining shuttle finally works on the Aetherwhisp and the Scarab. We did it reddit.
del: The main ship can no longer land on planets.
/🆑


Qustinnus:
why didnt you just push to my branch

ike709:
@Qustinnus I said it in discord nerd

ike709:
image

code/controllers/subsystem/starmap.dm
```diff
@@ -341,7 +341,7 @@ SUBSYSTEM_DEF(starmap)
 	ftl_drive.status_update(message)
 
 /datum/controller/subsystem/starmap/proc/ftl_sound(var/sound) //simple proc to play a sound to the crew aboard the ship, also since I want to use minor_announce for the FTL notice but that doesn't support sound
-	for(var/area/shuttle/ftl/F in world)
+	for(var/area/shuttle/ftl/F in world - /area/shuttle/ftl/cargo/mining)
```
ninjanomnom: check if the shuttle is docked or not before excluding it, or check in the area if it's currently on the same z level as the ship.

code/datums/planet_loader/planet_loader.dm
```diff
@@ -37,6 +37,9 @@
 			PL.name_dock(D, D.encounter_type, params)
 			if(D.encounter_type == "main")
 				PL.main_dock = D
+			if(D.planet_dock)
+				var/obj/docking_port/stationary/fob_land/F = new(D.loc)
+				PL.docks |= F
```
ninjanomnom: This code creates 2 docks

code/datums/stars.dm
```diff
@@ -101,6 +101,11 @@
 			no_unload_reason = "RELAY"
 			return 0
 
+	for(var/obj/machinery/computer/shuttle/fob/F in GLOB.machines)
+		if((F.z in z_levels))
+			no_unload_reason = "FOB SHUTTLE"
+			return 0
```
ninjanomnom: instead of this make the fob landing activate keep_loaded

code/game/area/Space_Station_13_areas.dm
```diff
@@ -529,6 +529,7 @@ NOTE: there are two lists of areas in the end of this file: centcom and station
 /area/shuttle/ftl/cargo/mining
 	name = "Mining"
 	icon_state = "mining"
+	has_gravity = TRUE //The FOB shuttle has sci fi magic gravity
```
ninjanomnom: why?

code/modules/shuttle/computer.dm
```diff
@@ -24,13 +24,19 @@
 	var/list/options = params2list(possible_destinations)
 	var/obj/docking_port/mobile/M = SSshuttle.getShuttle(shuttleId)
 	var/dat = "Status: [M ? M.getStatusText() : "*Missing*"]<br><br>"
-	if(M)
+	if(istype(src, /obj/machinery/computer/shuttle/fob) && SSstarmap.in_transit || SSstarmap.in_transit_planet)
+		dat += "<B>Main ship is currently in transit.</B><br>"
+	else if(M && M.mode == SHUTTLE_TRANSIT)
+		dat += "<B>The shuttle is currently in transit.</B><br>"
```
ninjanomnom: this is a gross breach of oop practices. Use inheritance

code/modules/shuttle/ftl.dm
```diff
@@ -48,6 +48,46 @@
...
+	name = "FTL FOB"
+	id = "fob"
+	callTime = 650
+	preferred_direction = SOUTH
+	area_type = /area/shuttle/ftl/cargo/mining
+	dir = NORTH
+	dwidth = 15
+	dheight = 0
+	width = 22
+	height = 13
```
ninjanomnom: add new values to the json for the fob and make the docks use them so mappers can have different sizes

code/modules/shuttle/ftl.dm
```diff
@@ -138,6 +178,8 @@
 			var/list/ports_list = list()
 			data["ports"] = ports_list
 			for(var/obj/docking_port/stationary/D in SSstarmap.current_planet.docks)
+				if(D.planet_dock || D.name == "FOB Landing Zone") //Only the FOB can land on planets now
+					continue
```
ninjanomnom: just remove planet landing functionality instead of this which will leave the button in the ui

code/controllers/subsystem/starmap.dm
ninjanomnom: do the second actualy

code/game/area/Space_Station_13_areas.dm
ike709: Otherwise it doesn't have gravity in transit space. Unless it's remapped to have a grav gen on the Aetherwhisp and Scarab, this is it.

code/modules/shuttle/ftl.dm
ike709: Standard docking port practice is to var edit these values in a map editor. This is just a default since Aetherwhisp is the default map.

ninjanomnom: you'll still need to add fob shuttle sizes to the json for maps which have different size shuttles

code/game/area/Space_Station_13_areas.dm
ninjanomnom: Yes having no gravity while in transit is fine behavior for something with no gravity gen

code/datums/planet_loader/planet_loader.dm
ike709: Oops. Forgot about this when I was dealing with the two planet dock problem.

code/datums/stars.dm
ike709: no_unload_reason is displayed on the starmap. This indicates where the fob shuttle is. See a few lines above where tcomms relays do the same thing.

code/controllers/subsystem/starmap.dm
ike709: Done but my method is probably equally terrible.

code/modules/shuttle/ftl.dm
ike709: Old code from the two planet docks era. Woops.

code/datums/stars.dm
ike709: Although I should make it not show on the helms console if the FOB is docked with the main ship still.

code/modules/shuttle/computer.dm
ike709: Resolved, hopefully.

code/modules/shuttle/ftl.dm
ike709: bleh, fine

code/controllers/subsystem/starmap.dm
```diff
@@ -341,16 +341,21 @@ SUBSYSTEM_DEF(starmap)
 	ftl_drive.status_update(message)
 
 /datum/controller/subsystem/starmap/proc/ftl_sound(var/sound) //simple proc to play a sound to the crew aboard the ship, also since I want to use minor_announce for the FTL notice but that doesn't support sound
-	for(var/area/shuttle/ftl/F in world)
-		F << sound
+	var/area/shuttle/ftl/cargo/mining/M
+	if(M.z == ZLEVEL_STATION)	//check if the FOB shuttle is with the main ship
```
ninjanomnom: M is null here, it hasn't been assigned a value. Loop through the areas in GLOB.sortedAreas first then get a turf per iteration and check it's z

code/datums/map_config.dm
```diff
@@ -29,6 +29,13 @@
 	var/ftl_ship_dheight = 62
 	var/ftl_ship_width = 73
 	var/ftl_ship_height = 113
+
+	var/fob_ship_dir = NORTH
+	var/fob_ship_dwidth = 15
+	var/fob_ship_dheight = 0
+	var/fob_ship_width = 22
+	var/fob_ship_height = 13
+
```
ninjanomnom: This is the aetherwhisp's shuttle right?

code/datums/stars.dm
```diff
@@ -101,6 +101,11 @@
 			no_unload_reason = "RELAY"
 			return 0
 
+	for(var/obj/machinery/computer/shuttle/fob/F in GLOB.machines)
+		if((F.z in z_levels) && F.z != ZLEVEL_STATION)
+			no_unload_reason = "FOB SHUTTLE"
+			return 0
```
ninjanomnom: no, you remove this code entirely from here and set the unload reason when the fob lands and unset and unload when it leaves.

code/modules/shuttle/ftl.dm
```diff
@@ -48,6 +48,56 @@
...
+	height = SSmapping.config.fob_ship_height
+
+/obj/machinery/computer/shuttle/fob
+	name = "FOB shuttle console"
+	desc = "Used to call and send the FOB shuttle."
+	shuttleId = "fob"
+	possible_destinations = "fob_dock;fob_land"
+	no_destination_swap = 1
+	can_move_if_ship_moving = FALSE
+
```
ninjanomnom: switch all the New() to Initialize()

code/datums/map_config.dm
ike709: Yes. But I should rename fob_ship_ to fob_shuttle_

code/modules/shuttle/ftl.dm
ike709: I'm dumb.

code/controllers/subsystem/starmap.dm
```diff
@@ -341,19 +339,23 @@ SUBSYSTEM_DEF(starmap)
 	ftl_drive.status_update(message)
 
 /datum/controller/subsystem/starmap/proc/ftl_sound(var/sound) //simple proc to play a sound to the crew aboard the ship, also since I want to use minor_announce for the FTL notice but that doesn't support sound
-	for(var/area/shuttle/ftl/F in world)
-		F << sound
+	for(var/area/A in GLOB.sortedAreas)
+		var/turf/T = A.contents[1]
+		if(T.z == ZLEVEL_STATION)
+			A << sound
```
ninjanomnom: Make the loop typeless.

Also you need to be checking if it's an area on the ship, not just an area on the same z level.

Also did you even test this? Area contents aren't always turfs, though they are atoms.

code/controllers/subsystem/starmap.dm
```diff
@@ -341,19 +339,23 @@ SUBSYSTEM_DEF(starmap)
...
 	ftl_is_spooling = 0
 
 /datum/controller/subsystem/starmap/proc/ftl_rumble(var/message)
-	for(var/area/shuttle/ftl/F in world)
-		for(var/mob/M in F)
-			to_chat(M, "<font color=red><i>The ship's deck starts to shudder violently as the FTL drive begins to activate.</font></i>")
-			rumble_camera(M,150,12)
+	for(var/area/A in GLOB.sortedAreas)
+		var/turf/T = A.contents[1]
+		if(T.z == ZLEVEL_STATION)
```
ninjanomnom: Same thing here, make this typeless and check more than just z

code/modules/shuttle/on_move.dm
```diff
@@ -396,10 +396,32 @@ All ShuttleMove procs go here
...
+/obj/docking_port/mobile/fob/onShuttleMove(turf/newT, turf/oldT, rotation, list/movement_force, move_dir, old_dock)
+	. = ..()
+
+	//Unload the planet
+	var/obj/docking_port/stationary/S = old_dock
+	for(var/datum/star_system/SYS in SSstarmap.star_systems)
+		var/datum/planet/PL = SYS.get_planet_for_z(oldT.z)
+		if(S.name == "FOB Landing Zone" && SSstarmap.current_planet.name != PL.name)
+			PL.no_unload_reason = ""
+			PL.do_unload()
```
ninjanomnom: On second thought I'm not sure about unloading immediately, I'll need to check on this later if do_unload is check_ticked.

ninjanomnom:
You don't need to have me in the changelog

code/controllers/subsystem/starmap.dm
ike709: I did test it and it works.

code/modules/shuttle/on_move.dm
ike709: So what do you want me to do then?

ninjanomnom: comment it out for now and leave that bit to another pr. It'll be unloaded anyway if the ship moves.

TMTIME:
Time for a fun bug. tldr the zlevel the shuttle was on deleted itself when the ship jumped
Round 487

Relay was on
So the shuttle was landed at the starting planet, dug up sone dirt ect ect.
Went back to the ship, we jumped to another system. Went down in the shuttle and landed on the exact same planet.
So we think whatever its chaos up there lets keep mining.
The ship jumps and we think nothing of it because both the relay and the FOB shuttle keeps zones loaded.
And it just gets deleted.
Not too sure what else to say

Super3222:
Lighting on planets is also broken whenever you land on them.

code/modules/shuttle/on_move.dm
```diff
@@ -396,10 +396,34 @@ All ShuttleMove procs go here
...
+
+	//Unload the planet
+	var/obj/docking_port/stationary/S = old_dock
+	for(var/datum/star_system/SYS in SSstarmap.star_systems)
+		var/datum/planet/PL = SYS.get_planet_for_z(oldT.z)
+		if(S.name == "FOB Landing Zone" && SSstarmap.current_planet.name != PL.name)
+			PL.no_unload_reason = ""
+			//PL.do_unload()
+	for(var/mob/living/simple_animal/M in range(30, S.loc)) //Stops the shuttle from landing on top of mobs
+		qdel(M)
```
ninjanomnom: you don't need this, shuttles innately handle landing on mobs though I think I have some notes somewhere about needing to tweak it a bit

code/modules/shuttle/on_move.dm
```diff
@@ -396,10 +396,34 @@ All ShuttleMove procs go here
...
 
+/obj/docking_port/mobile/fob/onShuttleMove(turf/newT, turf/oldT, rotation, list/movement_force, move_dir, old_dock)
+	. = ..()
+
+	//Unload the planet
+	var/obj/docking_port/stationary/S = old_dock
+	for(var/datum/star_system/SYS in SSstarmap.star_systems)
+		var/datum/planet/PL = SYS.get_planet_for_z(oldT.z)
+		if(S.name == "FOB Landing Zone" && SSstarmap.current_planet.name != PL.name)
+			PL.no_unload_reason = ""
```
ninjanomnom: Check name first before looping through every system in the game. 

Make it a typeless loop.

You don't need to check where the planet is. If the no unload reason was set by you on landing you must remove it when you leave.

Preferably put this in beforeShuttleMove

code/modules/shuttle/on_move.dm
ninjanomnom: your use of range() here hurts me

ninjanomnom:
Just for clarity by the way, I have before/afterShuttleMove slated for some changes to make dock have better performance in the future. Put anything in there if you don't particularly care about it happening *exactly* when everything moves.

code/modules/shuttle/on_move.dm
ike709: They gib whatever they land on. In the case of necropolis tendrils, this will create chasms inside the shuttle.

ike709: Also I *did* ping you on discord to see if there was a better way to do this.

code/modules/mob/living/simple_animal/hostile/mining_mobs/necropolis_tendril.dm
```diff
@@ -26,6 +26,10 @@
 	del_on_death = 1
 	var/gps = null
 	var/obj/effect/light_emitter/tendril/emitted_light
+	var/spawn_loot = TRUE
+
+/mob/living/simple_animal/hostile/spawner/lavaland/gib()
+	spawn_loot = FALSE
```
ninjanomnom: you forgot to call the parent

code/modules/shuttle/computer.dm
```diff
@@ -36,7 +36,7 @@
 				continue
 			if(!M.check_dock(S))
 				continue
-			if(M.id == "fob" && M.getDockedId() != "fob_dock" && S.id != "fob_dock") //If the FOB isn't at the main ship, the main ship is its only potential destination
+			if(M.id == "fob" && M.getDockedId() != "fob_dock" && S.id != "fob_dock" && (S.z == ZLEVEL_STATION || S.z in SSstarmap.current_planet.z_levels)) //If the FOB isn't at the main ship, the main ship is its only potential destination. Check the z level too.
```
ninjanomnom: split this up a bit for readability's sake. It's alright to have multiple if checks for continue.

code/modules/shuttle/on_move.dm
ninjanomnom: You didn't address all the issues

code/modules/shuttle/on_move.dm
```diff
@@ -401,13 +401,14 @@ All ShuttleMove procs go here
 
 	//Unload the planet
 	var/obj/docking_port/stationary/S = old_dock
-	for(var/datum/star_system/SYS in SSstarmap.star_systems)
-		var/datum/planet/PL = SYS.get_planet_for_z(oldT.z)
-		if(S.name == "FOB Landing Zone" && SSstarmap.current_planet.name != PL.name)
+	if(S.name != "FOB Landing Zone")
+		return
+	for(var/SYS in SSstarmap.star_systems)
```
ninjanomnom: I still really hate this loop through all star_systems. Store the star system in the fob dock instead.

ninjanomnom: Also, the previous review of this code:

~~Check name first before looping through every system in the game.~~

~~Make it a typeless loop.~~

You don't need to check where the planet is. If the no unload reason was set by you on landing you must remove it when you leave.

Preferably put this in beforeShuttleMove

code/modules/mob/living/simple_animal/hostile/mining_mobs/necropolis_tendril.dm
ike709: Damn it.

code/modules/shuttle/on_move.dm
```diff
@@ -393,13 +393,34 @@ All ShuttleMove procs go here
 /obj/docking_port/stationary/onShuttleMove(turf/newT, turf/oldT, rotation, list/movement_force, move_dir, old_dock)
 	if(old_dock == src) //Don't move the dock we're leaving
 		return FALSE
+	. = ..()
 
+/obj/docking_port/mobile/fob/beforeShuttleMove(turf/newT, rotation, move_mode)
+	src.previous_dock = src.get_docked()
```
ninjanomnom: You don't need `src.` here. 

`. = ..()` comes first

code/modules/shuttle/on_move.dm
```diff
@@ -393,13 +393,34 @@ All ShuttleMove procs go here
...
 	. = ..()
 
+/obj/docking_port/mobile/fob/afterShuttleMove(list/movement_force, shuttle_dir, shuttle_preferred_direction, move_dir)
+	. = ..()
+	//Keep the planet loaded
+	var/obj/docking_port/mobile/fob/M = src
+	var/obj/docking_port/stationary/fob_land/LZ = M.get_docked()
```
ninjanomnom: Stop using all caps names if they're longer than 1 char and not a define. Use something descriptive, that was half the point of the dock rewrite.

code/modules/shuttle/on_move.dm
```diff
@@ -393,13 +393,34 @@ All ShuttleMove procs go here
...
+	if(LZ.name == "FOB Landing Zone")
+		var/datum/planet/PL = LZ.current_planet
+		PL.no_unload_reason = "FOB SHUTTLE"
+
+	//Unload the planet
+	var/obj/docking_port/stationary/fob_land/OLD = M.previous_dock
```
ninjanomnom: Same as above, write self documenting code as best you can.

code/modules/shuttle/on_move.dm
```diff
@@ -393,13 +393,34 @@ All ShuttleMove procs go here
...
+	if(OLD.name != "FOB Landing Zone")
+		return
+	var/datum/planet/PL_OLD = OLD.current_planet
```
ninjanomnom: Same as above.

code/modules/shuttle/on_move.dm
```diff
@@ -393,13 +393,34 @@ All ShuttleMove procs go here
...
+	. = ..()
 
+/obj/docking_port/mobile/fob/beforeShuttleMove(turf/newT, rotation, move_mode)
+	src.previous_dock = src.get_docked()
 	. = ..()
 
+/obj/docking_port/mobile/fob/afterShuttleMove(list/movement_force, shuttle_dir, shuttle_preferred_direction, move_dir)
+	. = ..()
+	//Keep the planet loaded
+	var/obj/docking_port/mobile/fob/M = src
```
ninjanomnom: Why are you typecasting src?

code/modules/shuttle/ftl.dm
```diff
@@ -48,6 +48,59 @@
...
+	area_type = /area/shuttle/ftl/cargo/mining
+	var/previous_dock
+
+/obj/docking_port/mobile/fob/Initialize(mapload)
+	dir = SSmapping.config.fob_shuttle_dir
+	dwidth = SSmapping.config.fob_shuttle_dwidth
+	dheight = SSmapping.config.fob_shuttle_dheight
+	width = SSmapping.config.fob_shuttle_width
+	height = SSmapping.config.fob_shuttle_height
+	. = ..()
```
ninjanomnom: This is the first time I actually looked at this bit of code, make `/obj/docking_port/mobile/fob` the parent of everything else so they share an Initialize()

code/modules/shuttle/on_move.dm
ike709: ninja we both know the reason is my stupidity

code/modules/shuttle/on_move.dm
```diff
@@ -393,13 +393,33 @@ All ShuttleMove procs go here
...
 	. = ..()
+	previous_dock = get_docked()
+
+/obj/docking_port/mobile/fob/afterShuttleMove(list/movement_force, shuttle_dir, shuttle_preferred_direction, move_dir)
+	. = ..()
+	//Keep the planet loaded
+	var/obj/docking_port/stationary/fob/fob_land/destination_dock = get_docked()
+	if(destination_dock.name == "FOB Landing Zone")
+		var/datum/planet/planet = destination_dock.current_planet
+		planet.no_unload_reason = "FOB SHUTTLE"
```
ninjanomnom: you can return after this block since you're obviously not leaving if you just arrived.

code/modules/shuttle/on_move.dm
```diff
@@ -393,13 +393,33 @@ All ShuttleMove procs go here
...
+
+	//Unload the planet
+	var/obj/docking_port/stationary/fob/fob_land/old_dock = previous_dock
+	if(old_dock.name != "FOB Landing Zone")
+		return
+	var/datum/planet/old_planet = old_dock.current_planet
+	if(SSstarmap.current_planet.name != old_planet.name)
+		old_planet.no_unload_reason = ""
```
ninjanomnom: for the third time now, this if check is unnecesary. Check that the no_unload_reason is the same as you set it and if it is then assume there's nothing else and set it blank.

I'd ask this to be remade into a bitflag but you'd cry out in agony at this point

ike709: And on the third time I finally understood what you meant.

And you're right I would cry out in agony.

PR #2546, opened 2017-09-25
Balance droids to be a little more scary

Funce:
At the request of @Ecolli, Droids are now a little more scary. 

:cl: Funce
balance: Droids can now withstand more than just space.
/:cl:


ninjanomnom:
Duplicate of #2519

PR #2578, opened 2017-09-27
Mirror fix [#1834}

Qustinnus:
Original PR: tgstation/tgstation#29666

🆑 Floyd
fix: No longer does everyone look like a sick, nauseated weirdo!
/🆑

Fixes #29672

I must've missed this when I changed the examine text to a switch from an if else stack.

fixes #1834


Qustinnus:
@AuroranAI i did it dad

PR #2590, opened 2017-09-28
Wall chargers

stobarico:
why don't we have those?


Qustinnus:
pics

stobarico:
6e2ffd0b1dedb5ebd293c12c038ed05d
Animated parts:
https://i.gyazo.com/5a55a0e55a99a0f6f578bd9e56155b82.mp4
https://i.gyazo.com/6e34a0ef4d7d4e02fffab3118a26a669.mp4




pics or it never happened

Qustinnus:
look cool

code/game/machinery/recharger.dm
```diff
@@ -163,3 +167,11 @@
...
 		return
 	icon_state = "recharger0"
+
+obj/machinery/recharger/wallcharger
+	name = "wall recharger"
+	icon_state = "wrecharger0"
+	icon_state_off = "wrechargeroff"
+	icon_state_idle = "wrecharger0"
+	icon_state_charging = "wrecharger1"
+	icon_state_charged = "wrecharger2"
```
ninjanomnom: Did you even test this in game? This isn't how icon_states work.

Also if rechargers are going to have subtypes, split their icons into a new file altogether.

ninjanomnom:
image

ninjanomnom:
What part of "if rechargers are going to have subtypes..." makes you think I want you to ignore all good sense and copy paste all the code into a new object type.

code/game/machinery/recharger.dm
```diff
@@ -163,3 +168,90 @@
...
+		power_change()
+		to_chat(user, "<span class='notice'>You [anchored ? "attached" : "detached"] [src].</span>")
+		playsound(loc, G.usesound, 75, 1)
+		return
+
+	var/allowed = is_type_in_typecache(G, allowed_devices)
+
+	if(allowed)
+		if(anchored)
+			if(charging || panel_open)
```
ninjanomnom: avoid deep nesting like this and do stuff like `if(x) {return}`

code/game/machinery/recharger.dm
```diff
@@ -163,3 +168,90 @@
...
+				return 1
+
+			//Checks to make sure he's not in space doing it, and that the area got proper power.
+			var/area/a = get_area(src)
+			if(!isarea(a) || a.power_equip == 0)
```
ninjanomnom: space has an area

code/game/machinery/recharger.dm
```diff
@@ -10,6 +10,11 @@
 	var/obj/item/charging = null
 	var/static/list/allowed_devices = typecacheof(list(/obj/item/weapon/gun/energy, /obj/item/weapon/melee/baton, /obj/item/ammo_box/magazine/recharge, /obj/item/device/modular_computer))
 	var/recharge_coeff = 1
+	var/icon_state_off = "rechargeroff"
+	var/icon_state_charged = "recharger2"
+	var/icon_state_charging = "recharger1"
+	var/icon_state_idle = "recharger0"
```
ninjanomnom: These aren't used anywhere and you shouldn't add new vars for this sort of thing anyway

code/game/machinery/recharger.dm
```diff
@@ -163,3 +168,90 @@
...
+	name = "wall recharger"
+	icon = 'icons/obj/wallcharger.dmi'
+	icon_state = "wrecharger0"
+
+/obj/machinery/recharger/wallrecharger/Initialize()
+	. = ..()
+	var/obj/item/weapon/circuitboard/machine/recharger/wallrecharger/B = new()
+	B.apply_default_parts(src)
+
+/obj/item/weapon/circuitboard/machine/recharger/wallrecharger
```
ninjanomnom: This would only be necessary if you're making it constructable.

code/game/machinery/recharger.dm
```diff
@@ -163,3 +168,90 @@
 			icon_state = "recharger2"
 		return
 	icon_state = "recharger0"
+
+/obj/machinery/recharger/wallrecharger
+	name = "wall recharger"
+	icon = 'icons/obj/wallcharger.dmi'
```
ninjanomnom: Since you've moved both charger icons to a new file this needs to be changed in the parent, not here

code/game/machinery/recharger.dm
```diff
@@ -163,3 +168,90 @@
...
+/obj/machinery/recharger/wallrecharger/Initialize()
+	. = ..()
+	var/obj/item/weapon/circuitboard/machine/recharger/wallrecharger/B = new()
+	B.apply_default_parts(src)
+
+/obj/item/weapon/circuitboard/machine/recharger/wallrecharger
+	name = "Weapon Wall Recharger (Machine Board)"
+	build_path = /obj/machinery/recharger/wallrecharger
+	origin_tech = "powerstorage=4;engineering=3;materials=4"
+	req_components = list(/obj/item/weapon/stock_parts/capacitor = 1)
```
ninjanomnom: If you actually want to make it constructable have a second component at least. A matter bin would work.

Also move the construction stuff to the parent

code/game/machinery/recharger.dm
```diff
@@ -163,3 +168,90 @@
...
+			//Checks to make sure he's not in space doing it, and that the area got proper power.
+			var/area/a = get_area(src)
+			if(!isarea(a) || a.power_equip == 0)
+				to_chat(user, "<span class='notice'>[src] blinks red as you try to insert [G].</span>")
+				return 1
+
+			if (istype(G, /obj/item/weapon/gun/energy))
+				var/obj/item/weapon/gun/energy/E = G
+				if(!E.can_charge)
+					to_chat(user, "<span class='notice'>Your gun has no external power connector.</span>")
```
ninjanomnom: Why are you just duplicating code from the parent

code/game/machinery/recharger.dm
stobarico: I created a different file for the wall rechargers, not just the rechargers

code/game/machinery/recharger.dm
stobarico: I am making it constructable, and i tested it to be sure it worked and was fully constructable/deconstructable

code/game/machinery/recharger.dm
stobarico: Why? It's literally a charger on the wall, it should only take a different board

code/game/machinery/recharger.dm
stobarico: Simply copied from the recharger code.

code/game/machinery/recharger.dm
stobarico: It's from the recharger code

code/game/machinery/recharger.dm
stobarico: It wasn't working without that

code/game/machinery/recharger.dm
ninjanomnom: Doesn't matter, code improvement is always in scope

code/game/machinery/recharger.dm
ninjanomnom: Doesn't matter, code improvement is always in scope

code/game/machinery/recharger.dm
ninjanomnom: Then you'll have to either modify the parent or partialy override. Either way full scale copy pasting like this isn't allowed.

code/game/machinery/recharger.dm
ninjanomnom: Then move the regular ones as well.

code/game/machinery/recharger.dm
ninjanomnom: They would both use the same board but you screwdriver the board to switch modes like a few other machines (freezer/heater for example)

code/game/machinery/recharger.dm
stobarico: It isn't supposed to return if the weapon is allowed in it, it's supposed to go foreward and check if it's anchored

code/game/machinery/recharger.dm
stobarico: If that line is removed, the recharger can be used by any weapon, that means even the ones without a power connection (Like the advanced energy gun)

code/game/machinery/recharger.dm
stobarico: The recharger doesn't work on space

ninjanomnom:
\>using github to change files

stobarico:
They won't be needed anymore

PR #2684, opened 2017-10-02
Re-balances shields and moves some files to an FTL submodule for modularization

Qustinnus:
:cl: Floyd
tweak: Shieldgen has now been rebalanced, it has become more useful, but also more draining
/:cl:

Re-does some shit from my old combat PR but does it less shittily, you passively gain 200 shield power per process of the shieldgen, 4K is the default max. Plasma charges the same. If you get hit and your power is below 500, all the plasma is drained from the shields, (As if the power was holding the plasma in place)

From now on, all code changes should be done in the FTL submodule, and everything should be placed in the same typepath (similar to what I did). This allows us to prevent conflicts etc and make our lives easier.


code/modules/ftl/code/game/machinery/ftl_shieldgen.dm
```diff
@@ -13,25 +13,32 @@
...
+	var/charging_plasma = FALSE
+	var/charging_power = FALSE
+
+	var/load = 50000 //Power needed to charge the shield
+
+	var/minimal_power_charge = 500
+	var/minimal_plasma_charge = 50
+
+	var/power_charge_loss = 1000 //Amount of power lost per hit
+	var/power_charge_rate = 200 //Amount of shield gained per process11
```
AuroranAI: 1111111

code/controllers/subsystem/ships.dm
```diff
@@ -133,15 +133,18 @@ SUBSYSTEM_DEF(ship)
...
 		if(SSstarmap.ftl_shieldgen && SSstarmap.ftl_shieldgen.is_active() && !W.attack_data.shield_bust)
-			SSstarmap.ftl_shieldgen.take_hit()
-			broadcast_message("<span class=warning>Enemy ship ([S.name]) fired their [W.name] and hit! Hit absorbed by shields.",error_sound,S)
-			for(var/area/shuttle/ftl/A in world)
-				A << 'sound/weapons/ship_hit_shields.ogg'
+			if(W.attack_data.shield_damage)
+				SSstarmap.ftl_shieldgen.take_hit(W.attack_data.shield_damage)
+				broadcast_message("<span class=warning>Enemy ship ([S.name]) fired their [W.name] and hit! Hit absorbed by shields.",error_sound,S)
+				for(var/area/shuttle/ftl/A in world)
+					A << 'sound/weapons/ship_hit_shields.ogg'
```
ninjanomnom: use ftl_sound() instead

code/modules/ftl/code/game/machinery/ftl_shieldgen.dm
```diff
@@ -167,22 +170,17 @@
...
-			qdel(S)
-		shield_barrier_objs.Cut()
+	do_update = FALSE
+	while(shield_barrier_objs.len)
+		var/obj/to_remove = pick(shield_barrier_objs)
+		qdel(to_remove)
+		shield_barrier_objs -= to_remove
+		if(prob(10))
+			sleep(1)
+	do_update = TRUE
```
ninjanomnom: instead of using extra vars making sure this isn't called too much, make all calls to this proc a unique addtimer

Qustinnus: how would that work?

PR #2717, opened 2017-10-04
Supermatter awareness alert (TM)

Cdey78:
A.K.A:
Shit your pants because the CE almost delaminated the engine!

This makes it so when the supermatter delaminates, Delta alert will sound and dim, red emergency lights will activate. Now there is no doubt about when to call the shuttle!
(Does not require power for them to function)

:cl: Cdey
add: Supermatter delaminations now sound the Delta alert.
add: Also adds the "Emergency lights", which means when the Delta alert goes off, so do these! Warning: Basically just aesthetics, they're extremely dim and the chances of them helping in any way shape or form is VERY low.
/:cl:

image

image

Non-engineer-lives matter!

### DIFF
--- a/code/modules/power/lighting.dm
+++ b/code/modules/power/lighting.dm
@@ -221,7 +221,7 @@ GLOBAL_VAR_INIT(emergency_light, 0)
 					set_light(2, 0.5, "#ff0000")
 					icon_state = "firelight1"
 				else
-					set_light(0, 0.5, "#ff0000")
+					set_light(0, 0, "#ff0000")
 					icon_state = "firelight0"
 		spawn(1)
 			update(0)
@@ -234,20 +234,22 @@ GLOBAL_VAR_INIT(emergency_light, 0)
 	return ..()
 
 /obj/machinery/light/update_icon()
-
-	switch(status)		// set icon_states
-		if(LIGHT_OK)
-			icon_state = "[base_state][on]"
-		if(LIGHT_EMPTY)
-			icon_state = "[base_state]-empty"
-			on = FALSE
-		if(LIGHT_BURNED)
-			icon_state = "[base_state]-burned"
-			on = FALSE
-		if(LIGHT_BROKEN)
-			icon_state = "[base_state]-broken"
-			on = FALSE
-	return
+	if(special == 1)
+		return
+	else
+		switch(status)		// set icon_states
+			if(LIGHT_OK)
+				icon_state = "[base_state][on]"
+			if(LIGHT_EMPTY)
+				icon_state = "[base_state]-empty"
+				on = FALSE
+			if(LIGHT_BURNED)
+				icon_state = "[base_state]-burned"
+				on = FALSE
+			if(LIGHT_BROKEN)
+				icon_state = "[base_state]-broken"
+				on = FALSE
+		return
 
 // update the icon_state and luminosity of the light depending on its state
 /obj/machinery/light/proc/update(trigger = 1)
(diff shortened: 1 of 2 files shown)


Cdey78:
I do plan on adding them into engineering in a separate PR, due to some.. Issues with a previous PR.

code/modules/power/lighting.dm
```diff
@@ -225,49 +234,59 @@
...
-			icon_state = "[base_state]-empty"
-			on = FALSE
-		if(LIGHT_BURNED)
-			icon_state = "[base_state]-burned"
-			on = FALSE
-		if(LIGHT_BROKEN)
-			icon_state = "[base_state]-broken"
-			on = FALSE
-	return
+	if(special == 1)
```
Funce: Special conditions, like this, is called 'Hacky Code' and generally isn't allowed anywhere and isn't good practice. Overload this proc with your emergency lights instead.

code/modules/power/lighting.dm
```diff
@@ -225,49 +234,59 @@
...
-			if(rigged)
-				if(status == LIGHT_OK && trigger)
-					explode()
-			else if( prob( min(60, switchcount*switchcount*0.01) ) )
-				if(trigger)
-					burn_out()
-			else
-				use_power = ACTIVE_POWER_USE
-				set_light(brightness)
+	if(special == 1)
```
Funce: As above/before.

code/modules/power/supermatter/supermatter.dm
```diff
@@ -196,6 +203,31 @@
...
+/obj/machinery/power/supermatter_shard/proc/secure_alert()
+	if(alert_set == 1)
+		set_security_level(previous_level)
+		alert_set = 0
+		GLOB.emergency_light = 0
+		update_light()
+		for(var/obj/machinery/light/small/emergency/E in GLOB.machines)
+			E.update()
+	else
+		return
```
ninjanomnom: You don't need any of the returns

code/modules/power/lighting.dm
```diff
@@ -269,6 +278,15 @@
...
 			removeStaticPower(static_power_used, STATIC_LIGHT)
 
+/obj/machinery/light/small/emergency/update(trigger = 1)
+	update_icon()
+	if(GLOB.emergency_light == 1)
+		set_light(2, 0.5, "#ff0000")
+		icon_state = "firelight1"
+	else
+		set_light(0, 0, "#ff0000")
+		icon_state = "firelight0"
```
ninjanomnom: Use inheritance, don't copy paste

code/modules/power/lighting.dm
```diff
@@ -214,6 +215,13 @@
 				brightness = 4
 				if(prob(5))
 					break_light_tube(1)
+			if("emergency")
+				if(GLOB.emergency_light == 1)
+					set_light(2, 0.5, "#ff0000")
+					icon_state = "firelight1"
+				else
+					set_light(0, 0, "#ff0000")
+					icon_state = "firelight0"
```
ninjanomnom: Put this in update not new

code/modules/power/lighting.dm
```diff
@@ -9,6 +9,7 @@
 #define LIGHT_BROKEN 2
 #define LIGHT_BURNED 3
 
+GLOBAL_VAR_INIT(emergency_light, 0)
```
ninjanomnom: Instead of a new global var make an actual alert for this or use an existing one for these emergency lights

code/modules/power/supermatter/supermatter.dm
```diff
@@ -196,6 +203,31 @@
 		return SUPERMATTER_NORMAL
 	return SUPERMATTER_INACTIVE
 
+/obj/machinery/power/supermatter_shard/proc/delta_alert()
```
ninjanomnom: just make this a single set_alert proc

Cdey78:
Fixed those, I believe.

ninjanomnom:
I'd honestly be surprised this even works in game.

Vivalas:
Hm, do we have colored lighting now?

In that case I think it's best the emergency lights just be dropped completely and this behavior be added to fire alarms instead. In fact, I was planning on doing this anyways soon, so I might get to it.

crossedfall:
Stale- might be something we can add in 2.0.

PR #2827, opened 2017-10-08
Fixes the ready buttons not updating

Funce:
This was kind of annoying me

:cl: Funce
fix: Ready Buttons properly show if you're ready
/:cl:


ninjanomnom:
This shouldn't make a difference o.o

Did you check in game if it fixed it?

Funce:
Strangely enough, this was the only thing that didn't match up quite right.

This fixes the bug strangely and I'm not sure why. Maybe a secret refresh topic call is there

Funce:
Yes this fixes it.

PR #2832, opened 2017-10-08
Potential fix for the shuttle being unloaded.

### LINKED ISSUE #2825: Mining Shuttle despawning
[Note]: # (<B>Please enter the `Round ID` from the `Status` panel if you can!</B> If you believe the issue to be caused by a testmerge, please report it in its relative PR thread. State what the issue is from a "whats wrong" prospective. Issue reports should clearly allow maintainers to understand whats wrong and how to test/reproduce if that is not obvious. Avoid ambiguity. Start your issue report below both of these lines (or remove them))

[Admins]: # (If you are reporting a bug that occured AFTER you used varedit/admin buttons to alter an object out of normal operating conditions, please verify that you can re-create the bug without the varedit usage/admin buttons before reporting the issue.)

Round ID 534 - possibly 533 as well. miningshuttle failed to protect us from planet overwrite.

ike709:
:cl: ike709
fix: Potentially fixed the fob shuttle's planet being unloaded. Let me know if that still happens.
/:cl:

Potentially fixes #2825. I'm pretty sure the relay is overriding `no_unload_reason = "FOB SHUTTLE"` resulting in the planet being unloaded if the relay is turned off.


code/datums/stars.dm
```diff
@@ -97,7 +97,7 @@
```
Qustinnus: More likely: if(!main_dock) is somehow causing this, move this down to the bottom so it checks FOB_SHUTTLE first. if RELAY is removed, it would just stop at FOB_SHUTTLE anyways

PR #3058, opened 2017-10-17
Ports sounds

matskuman5:
Ports some sounds from Yog

:cl: 
soundadd: Ports miscellanious sounds from Yogstation
/:cl:

[why]: 
bcuz sounds r kewl

PR #3116, opened 2017-10-20
Makes buttons go "click" when pressed

stobarico:
Not really much else to say about this
It has been tested, it does indeed make a "click"

[button.zip](https://github.com/FTL13/FTL13/files/1403357/button.zip)
it's in a ZIP because it can't take audio files


ninjanomnom:
#3058 Does this same thing and adds more sounds

PR #3238, opened 2017-10-28
Makes the ship subsystem tick and sets it as a background system

ninjanomnom:
@Vivalas did you not read any other subsystem code before making this or something?

There are other major issues but qdeleted ship runtimes should be less frequent and this also means the ship subsystem won't hold the rest of the game hostage when it starts lagging.


Vivalas:
Eh, I didn't really know what to look at and it seemed to work fine, do you want me to merge this?

PR #3448, opened 2017-11-23
Updates meteors, makes Meteor event less threatening 

ncaaseawolf:
Lowers the damage of most meteors, lowers the amount of meteors spawned per event.


LordWalnut:
Lowers? I think it should be higher!

Qustinnus:
why do you think this is necesary

PR #3452, opened 2017-12-17
Make the FOB locked message more clear

awfulworldkid:
:cl:tweak: tweaked a few things - Made it easier to tell why the shuttle is locked/:cl:

[why]: Because it makes it less rage inducing when you don't know that you're over a gas giant.


MrMaclang:
changelogs dosent work like that, here is a format for that to work
:cl:
tweak: tweaked a few things - Made it easier to tell why the shuttle is locked
/🆑

PR #3455, opened 2017-12-23
Performance improvements and slight restructuring to ship subsystem processing

ninjanomnom:
Lots of minor tweaks that should change no actual behaviors of ships like list caching and avoiding loops which duplicate the list.

Biggest gain here is:
`while(!C && find_broken_ship_components(S))`
Every tick, every ship, let's call a proc that checks if at least one component is broken _in a loop going through every component anyway_

### DIFF
--- a/code/controllers/subsystem/ships.dm
+++ b/code/controllers/subsystem/ships.dm
@@ -32,6 +32,7 @@ SUBSYSTEM_DEF(ship)
 	if(!resumed)
 		src.currentrun = ships.Copy()
 		process_factions()
+		SSstarmap.process_economy()
 
 	var/list/currentrun = src.currentrun
 	while(currentrun.len)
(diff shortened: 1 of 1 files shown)


ninjanomnom:
Travis is broken atm due to a dependency in the build tools being broken

PJB3005:
> Biggest gain here is:
`while(!C && find_broken_ship_components(S))`
Every tick, every ship, let's call a proc that checks if at least one component is broken in a loop going through every component anyway

I can't quite tell whether you're joking or not.

First of all, *yes*, this code is pretty horrible.

However, this doesn't loop over every component, and unless these ships have literally *millions* of components, there is no conceivable way there is any proc call overhead here like you claimed on Discord. Ever.

code/controllers/subsystem/ships.dm
```diff
@@ -117,24 +118,24 @@ SUBSYSTEM_DEF(ship)
...
-
-		while(!C && find_broken_ship_components(S)) //pick a broken ship_component to fix
-			C = pick(S.ship_components)
-			if(C.active) C = null
-		if(C)
-			C.active = 1 //fix that shit
-
+		var/list/ship_components = S.ship_components
+		var/list/broken_components = list()
+		for(var/i in 1 to ship_components.len)
```
PJB3005: Use `for (var/datum/ship_component/C in ship_components)` instead.

code/controllers/subsystem/ships.dm
```diff
@@ -117,24 +118,24 @@ SUBSYSTEM_DEF(ship)
...
 		S.next_repair = world.time + S.repair_time
-		var/datum/ship_component/C
-
-		while(!C && find_broken_ship_components(S)) //pick a broken ship_component to fix
-			C = pick(S.ship_components)
-			if(C.active) C = null
-		if(C)
-			C.active = 1 //fix that shit
-
+		var/list/ship_components = S.ship_components
```
PJB3005: Why even store this variable locally?

Oh I know because if it's in a subsystem it *has* to do the most dead horse beating you could possibly imagine.

code/controllers/subsystem/ships.dm
```diff
@@ -533,25 +530,28 @@ SUBSYSTEM_DEF(ship)
 	S.ftl_vector = target
 
 /datum/controller/subsystem/ship/proc/process_factions()
-	for(var/datum/star_faction/faction in star_factions)
+	var/list/star_factions = src.star_factions //cache for free performance
```
PJB3005: > free performance

oh fuck off the difference is so immeasurable you should be ashamed to even consider this.

monster860:
The only real way to get a performance boost in BYOND is to not use BYOND.

MrMaclang:
try the supermatter engine, better than BYOND engine

Dennok:
Why not  add component  to "broken_components" when player is total damage it?

PJB3005:
Yes that is totally the correct way this should've been optimized but IIRC
the code is hot spaghetti.

ninjanomnom:
That would indeed be better but this pr purposefully doesn't make any signature changes to the functions.

PR #3457, opened 2017-12-28
engine plasmaglass

MrMaclang:
:cl:
tweak: your engine uses reinforced plasmaglass instead of normal reinforced glass
fix: fixed air tanks directional issue
/:cl:

[why]: why are we not using plasmaglass??!1!


Cdey78:
It doesn’t use normal reinforced glass, it uses pressure-hardened glass.

Cdey78:
Check it’s type before trying to change it.

Gobblesmack:
the engine fix is for the aetherwhisp
the air tank fix is for the scarab

highlight which map it is in the changelog and also capitalize your changelog for the love of god 
thanks

Cdey78:
image

Also, you didn't map merge.

MrMaclang:
it is mapmerged reee

Cdey78:
image
You sure about that?

MrMaclang:
ye, go read the map .dmm if you want

Qustinnus:
mapmerge and fix conflicts

PR #3459, opened 2018-01-01
Drone dispenser preloaded for 3 drones

petermorck:
Change existing drone dispenser to the preloaded
variant, preloaded with materials for 3 drones.
Also change the max material storage to 20 sheets.

[Changelogs]:
:cl: Rymdmannen
tweak: Replaced drone dispenser with preloaded version that holds materials for 3 drones. Also changed the material storage to 20 sheets instead of 100.
/:cl:

[why]: Because nobody ever remembers putting material in them, and if they do they put 2 full stacks.


petermorck:
Urgh, forgot to change it on Scarab.

PR #3460, opened 2018-01-02
Drone dispenser preloaded for 3 drones

petermorck:
Change existing drone dispenser to the preloaded variant on
Aetherwhisp and Scarab, preloaded with materials for 3 drones.
Also reduce the max material storage to 20 sheets from 100.

:cl: Rymdmannen
tweak: Replaced drone dispenser on Aetherwhisp and Scarab with preloaded version that holds materials for 3 drones. Also changed the material storage to 20 sheets instead of 100.
/:cl:

[why]: Because nobody ever remembers putting material in them, and if they do they put 2 full stacks.

### DIFF
--- a/_maps/map_files/Aetherwhisp/Aetherwhisp.dmm
+++ b/_maps/map_files/Aetherwhisp/Aetherwhisp.dmm
@@ -15102,7 +15102,7 @@
 /turf/open/floor/plating,
 /area/shuttle/ftl/maintenance/medbay)
 "iWm" = (
-/obj/machinery/droneDispenser,
+/obj/machinery/droneDispenser/preloaded,
 /turf/open/floor/plasteel/purple/side,
 /area/shuttle/ftl/assembly/robotics)
 "iWt" = (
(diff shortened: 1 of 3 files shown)


petermorck:
According to @ninjanomnom this build failure is fixed on TG, I assume by this PR:
https://github.com/tgstation/tgstation/pull/33775

The-clown-known-as-Pepe:
bad, removes interdepartamental cooperation

petermorck:
On Aetherwhisp there are what, 100 sheets of metal in the **same room** as the dispenser, a stack in the room outside (R&D) and I think also another stack in the storage room just across the hallway.
So there are easily 200 stacks available at start within 16 tiles of walking from the dispenser, and using or getting to any of them doesn't require any interdepartamental cooperation, yet it rarely gets done.
The glass sheets are slightly fewer but aren't used as much on the other hand, so they're really not an issue either. Also, 3 drones correspond to 3 sheets total of the two most basic materials that you can get practically anywhere. It's not like it needs 10 diamonds and 10 sheets of bananium per drone.
The problem is that very often people simple do not put ANYTHING in the dispenser and just ignore it. Letting it spawn with materials for three lousy drones is hardly removing any cooperation because it's not needed to begin with.

Altangy:
There is already a preset for this in the map editor

PR #3461, opened 2018-01-13
NERFS FTL LAG (Makes planet gen occur when FTL starts, instead of at the end)

TMTIME:
an actual big fix holy heck

:cl: TMTIME
fix: Makes planetgen occur during FTL, instead of when the timer hits 0
tweak: Increased planet to planet jump time as 60 seconds isn't enough for the potato
/:cl:

[why]: # b u g 

tldr for a long time it was assumed that planet gen was loading every planet in the system, causing FTL to always overrun the 3 minute timer.
looking into what was actually going on the planets would only be unloaded/loaded when the FTL was supposed to end according to the timer.

Give or take the performance on the server, and how habitable planets take forever this will cut the time down for each jump significantly

~~speedmerge when~~

late edit: One drawback to this PR is now anyone off the ship when it jumps gets deleted right away. No three minute window to ahelp for help

### DIFF
--- a/code/controllers/subsystem/starmap.dm
+++ b/code/controllers/subsystem/starmap.dm
@@ -20,7 +20,7 @@ SUBSYSTEM_DEF(starmap)
 	var/datum/planet/to_planet
 	var/in_transit_planet // In transit between planets?
 
-	var/is_loading = 0 // 0 = not loading. 1 = loading/unloading new z levels. 2 = recently ran, waiting for she ship to leave FTL
+	var/is_loading = FTL_NOT_LOADING // Status of z level loading during FTL
 
 	var/obj/machinery/ftl_drive/ftl_drive
 	var/obj/machinery/ftl_shieldgen/ftl_shieldgen
@@ -121,12 +121,12 @@ SUBSYSTEM_DEF(starmap)
 	..()
 
 /datum/controller/subsystem/starmap/fire()
-	if(is_loading == 1 && world.time >= to_time)
+	if(is_loading == FTL_LOADING && world.time >= to_time)
 		to_time += 100
 
 	if(in_transit || in_transit_planet)
 		var/obj/docking_port/mobile/ftl/ftl = SSshuttle.getShuttle("ftl")
-		if(ftl.mode == SHUTTLE_TRANSIT && is_loading == 0 && world.time >= from_time + 50)
+		if(ftl.mode == SHUTTLE_TRANSIT && is_loading == FTL_NOT_LOADING && world.time >= from_time + 50)
 			if(in_transit)
 				SSmapping.load_planet(to_system.planets[1])
 			else if(in_transit_planet)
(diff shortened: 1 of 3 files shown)


code/controllers/subsystem/starmap.dm
```diff
@@ -20,7 +20,7 @@ SUBSYSTEM_DEF(starmap)
 	var/datum/planet/to_planet
 	var/in_transit_planet // In transit between planets?
 
-	var/is_loading = 0
+	var/is_loading = 0 // 0 = not loading. 1 = loading/unloading new z levels. 2 = recently ran, waiting for she ship to leave FTL
```
Qustinnus: Use defines for this
for example

#define FTL_NOT_LOADING 1
#define FTL_LOADING 2
#define FTL_RECENT_FTL 3

then undefine them when you stop using them (bottom of the file most likely) by doing

#undefine FTL_NOT_LOADING 1
#undefine FTL_LOADING 2
#undefine FTL_RECENT_FTL 3

code/controllers/subsystem/starmap.dm
```diff
@@ -121,45 +121,45 @@ SUBSYSTEM_DEF(starmap)
...
 
 /datum/controller/subsystem/starmap/fire()
-	if((in_transit || in_transit_planet) && world.time > to_time)
+	if(is_loading == 1 && world.time >= to_time)
+		to_time += 100
+
+	if(in_transit || in_transit_planet)
 		var/obj/docking_port/mobile/ftl/ftl = SSshuttle.getShuttle("ftl")
-		if(ftl.mode == SHUTTLE_TRANSIT && !is_loading)
+		if(ftl.mode == SHUTTLE_TRANSIT && is_loading == 0 && world.time >= from_time + 50)
```
Qustinnus: replace with define

code/controllers/subsystem/starmap.dm
```diff
@@ -121,45 +121,45 @@ SUBSYSTEM_DEF(starmap)
...
-
-			to_time += 300
-			return
-
 		if(in_transit)
 			current_system = to_system
 			current_system.visited = TRUE
 			current_planet = current_system.planets[1]
 		else if(in_transit_planet)
 			current_planet = to_planet
```
monster860: Shouldn't this be done only after docking at the final destiation?

code/controllers/subsystem/starmap.dm
```diff
@@ -121,45 +121,45 @@ SUBSYSTEM_DEF(starmap)
...
 
-		var/obj/docking_port/stationary/dest = current_planet.main_dock
-		ftl.mode = SHUTTLE_CALL
-		ftl.destination = dest
+		if(is_loading == FTL_DONE_LOADING && world.time >= to_time && (in_transit || in_transit_planet))
```
monster860: ` && (in_transit || in_transit_planet)` seems a bit redundant here.

PR #3464, opened 2018-01-21
Disables weather subsystem

Vivalas:
Just makes it so weather subsystem starts disabled, at least until someone actually changes the weather effects to something more thematic, like snow or something.

:cl: Vivalas

del: Removes ash storms

/:cl:

[why]: Because they are out of place and annoying.


The-clown-known-as-Pepe:
hugbox

PR #3465, opened 2018-01-21
Aetherwhisp Map Changes

IndusRobot:

:cl: IndusRobot
add: Added more firelocks in case of fire.
fix: Cargo mass driver and disposals blast door now opens to cargo staff only.
fix: Rats in maintenance have been moved to reduce chance of maintenance blackouts.
fix: Monkeys in genetics have been moved back to their cage.
fix: FTL and Shield drives are now restricted to engineers and atmos techs only. 
fix: Chapel has its own area in maintenance. 
/:cl:

Message me if something else needs to be changed, and it wasn't listed in this changelog. 

[why]: # (Please add a short description [on the next line] of why you think these changes would benefit the game. If you can't justify it in words, it might not be worth adding:)

### DIFF
--- a/_maps/map_files/Aetherwhisp/Aetherwhisp.dmm
+++ b/_maps/map_files/Aetherwhisp/Aetherwhisp.dmm
@@ -375,24 +375,6 @@
 	},
 /turf/open/floor/circuit,
 /area/shuttle/ftl/research/server)
-"acD" = (
-/obj/structure/table/wood,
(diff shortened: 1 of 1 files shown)


The-clown-known-as-Pepe:
>fix: Rats in maintenance have been moved to reduce chance of maintenance blackouts.

mfw people can't be arsed to replace 2 wires or one hit kill simple mobs

PR #3479, opened 2018-02-12
Lower role priority for the Head of Security at round start

crossedfall:
Move the Head of Security to one of the last command level positions to be filled where applicable. This will force people to fill more ship critical spots (starting with the CE) before taking on the HoS role.

🆑 Crossedfall
Modification: Moved Head of Security down a few lines in the command positions
/🆑

### DIFF
--- a/code/modules/jobs/jobs.dm
+++ b/code/modules/jobs/jobs.dm
@@ -1,10 +1,10 @@
 GLOBAL_LIST_INIT(command_positions, list(
 	"Captain",
 	"Executive Officer",
-	"Head of Security",
 	"Chief Engineer",
 	"Research Director",
 	"Chief Medical Officer",
+	"Head of Security",
 	"Bridge Officer",))
 
 


Cdey78:
CMO is more valuable than an RD..
ReeeeeeeeeeeeEEEeEEeEEEEEEeeeeeeee

crossedfall:
Not really since the RD does R&D and gets weapon upgrades and ions RRrrrreeeEEEeeEEEeeEEe

Cdey78:
Crew survival > Fucking wormhole gun

crossedfall:
Crew survival can be determined by the ship having properly upgraded weapons you wafflecake. If the ship has the basic kit it is pretty unlikely to last very many fights, it will run out of power way to often and the damage will be terribad.

Cdey78:
>unlikely to survive many fights
Yeah...
Except they can.
The literal only issue is the fucking Syndicate Capital.
Power will be fine if the engineers aren't idiots.
AND, if everyone's not retarded, they will replace the power cells with better ones from engineering.
The ship literally spawns with everything you will need.
The RD literally just gives them even more shit that they don't need.
Seriously, play a single round where there is no RnD, or xenobiology. It plays the exact same, only combat lasts slightly longer.

crossedfall:
I don't understand your complaints on this specific PR. All I did was move the Head of Security role down. The rest of the roles are left as their defaults.

Power from engineering has nothing to do with the power attribute for the gun, that's determined by the powercell. While true, the ship spawns with medium tier items for weapons, you cannot have higher tier upgrades without science (that being Super Capacity cells, pico manipulators, and so on). Additionally, it is the RD's job to ensure that the equipment is upgraded, having anyone else do it is a delegation of duty. You're more or less saying the same thing I am about the CMO, technically anyone can do it, but in terms of SOP the department should spearhead the task. That includes equipment upgrades, as I noted prior.

Last note, combat doesn't last longer with an RD it's much shorter. Additionally, ship damage tends to be heavily mitigated by the bridge-crew's ability to dispatch enemies more effectively. Less direct hits from enemies = lower likelihood of ship damage = lower chances of crew injury.

PR #3492, opened 2018-03-04
Adds a Hydrogen gas reaction

Koekiegast:
:cl: Koekiegast
add: Added a highly dangerous gas reaction to hydrogen, pure hydrogen not advised.
/:cl:

It's double the power of plasma because the element is so light it makes things reach the surface faster or something, anyway pure hydrogen is not advised. I tested it.

### DIFF
--- a/code/modules/power/supermatter/supermatter.dm
+++ b/code/modules/power/supermatter/supermatter.dm
@@ -6,9 +6,11 @@
 #define OXYGEN_HEAT_PENALTY 1
 #define CO2_HEAT_PENALTY 0.1
 #define NITROGEN_HEAT_MODIFIER -1.5
+#define HYDROGEN_HEAT_PENALTY 30
 
 #define OXYGEN_TRANSMIT_MODIFIER 1.5   //Higher == Bigger bonus to power generation.
 #define PLASMA_TRANSMIT_MODIFIER 4
+#define HYDROGEN_TRANSMIT_MODIFIER 8
 
 #define N2O_HEAT_RESISTANCE 6          //Higher == Gas makes the crystal more resistant against heat damage.
 
@@ -93,6 +95,7 @@
 	var/o2comp = 0
 	var/co2comp = 0
 	var/n2ocomp = 0
+	var/hydrogencomp = 0
 
 	var/combined_gas = 0
 	var/gasmix_power_ratio = 0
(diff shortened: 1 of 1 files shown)


code/modules/power/supermatter/supermatter.dm
```diff
@@ -6,9 +6,11 @@
 #define OXYGEN_HEAT_PENALTY 1
 #define CO2_HEAT_PENALTY 0.1
 #define NITROGEN_HEAT_MODIFIER -1.5
+#define HYDROGEN_HEAT_PENALTY 30
```
crossedfall: H2 shouldn't have a higher heat factor than plasma (let alone double the heat factor), it doesn't really make sense considering plasma's regarded physical properties. Maybe consider further nerfing plasma in parallel or reducing the overall nerf to hydrogen.

crossedfall:
There isn't anything here to address Hydrogen's unique characteristics. I would recommend adding in the unique characteristics to act as a proper counter balance rather than arbitrarily increasing its reaction with the shard specifically. This will add more dynamic balancing and will make handling feel a bit more natural

Some things to note about H2 as a gas:

- It is highly volatile when heated and exposed to air, most notably O2.
- It will explode at 4-75 percent by volume when exposed to a flame or spark
- Despite its stability, H2 forms loads of bonds

I think a good change here would be to implement a reaction to where at high concentration (maybe 45%+) and heated it will react quickly with the O2 created by the shard and start a fire. This fire **should** burn away the H2, however, leaving something else in its place (_maybe water vapor?_). That way it wont be toooo punishing, but will still provide some level of challenge. The extreme alternative is to use the high end of 60% and just make it explode if it gets hot and there is enough O2 in the chamber.

Koekiegast:
Well i can't put a video of the delam in here but it's not a fast one, perhaps i should test a n2o+hydrogen mix but i doubt that that would delam. So it isn't that bad.

Koekiegast:
Well with the exception of crossedfall i haven't had anyone want change, i'll probably adjust it a bit if i get hydrogen to do these suggested things but that's for another pr.

Qustinnus:
@Koekiegast get Altangy's opinion on this

Koekiegast:
Well basicly this was altangy's opinion

crossedfall:
To expand on some points I had prior...

Since the alterations seem arbitrary, my primary concern would be balance within a live working environment and the backlash from effectively removing most current "engine-metas," I would suggest performing the adjustments manually during rounds with competent engineers (after notifying them). This will help ensure that the player base can cope with the changes and to verify that the chosen values work within a live/working environment. 

The alternative will be heavily monitoring player interaction and frequently creating new PRs to adjust the balance until it’s at a level the staff is comfortable with. The Blizzard approach to balance, if you will. I would hamper a guess to say this would increase the workload on an already resource deprived team, however.

A final note, merging a balance pr where we are only adjusting gas values, without adding in gas properties, will create more work down the line; as the gas will need to be re-balanced anyways once the properties are added to ensure that it's not too punishing. In my opinion, I think it would be a little premature to adjust gas values before taking the inherit properties into consideration first. I would instead recommend creating a project targeted at reviewing the gases and adjusting them to provide a more consistent and reliable approach to rebalancing an already contentious system.

Koekiegast:
So after further conversation and testing i've decided to work on the other gas reactions first.

PR #3494, opened 2018-03-09
boarding but better

TMTIME:
~~stop avoiding boarding nerds~~

:cl: TMTIME
add: Boarding is now a ship objective!
balance: NPC crew on boarding maps. Will become more armed as player counts rise.
imageadd: blackbox_empty and tape reel
fix: Pirate ships now inherit pirate gear for defenders
/:cl:

[why]: 
okay what did i even do to get here:

NPCs are now on boarding maps. They won't attack defenders and start as simple civilians who just scream and run
As player count becomes higher, they have a better chance to upgrade to armed guards, capping at 75%
~~Sorry not sorry to the unlucky crew that gets 10 stormtroopers~~

Oh yeah objective boarding. The crew can now be assigned a boarding objective to board a ship.
The following changes occur if boarding is forced:

- No nuke, replaced with a ship blackbox
- No nuke timer. Take as long as you want
- ONLY that ship can be boarded in that system


Qustinnus:
seems ok to me

PR #3497, opened 2018-03-14
Half Lag 2: Episode 1

TMTIME:
i fukin did it i deleted the ship

:cl: TMTIME
tweak: Stopped moving the ship at the start/end of FTL
fix: This should fix cams
fix: The weird issue on local servers where the ship size isn't set right
/:cl:

[why]: # l a g

https://drive.google.com/file/d/1_ZWXX4ajlx1klrVJRJpa27KDDiNdR4aY/view

So this is not done ever but merge away for some sweet sweet lag-*less* FTL 

What this does do:

Stops moving the ship. ever.
Fakes the lights turning off with FTL, since that was kinda cool
Keeps knockdown
Sorta has transit space

Thing this maybe does?:
Fixes AI/xenobio cams since we never move the ship

Things this does not do:
Has the cool transit space effect
Things I'm not gonna list here bc they are funny

Also threw in a temp fix for ship sizes (the front of the ship missing) on local servers, but thats nor really needed now we don't move the station


code/controllers/subsystem/mapping.dm
```diff
@@ -140,6 +140,47 @@ SUBSYSTEM_DEF(mapping)
 		var/turf/open/floor/circuit/C = N
 		C.update_icon()
 
+/datum/controller/subsystem/mapping/proc/fake_ftl_change(var/ftl_start=TRUE)
+	var/turf/newspaceturf
+	var/throw_dir
+	var/flavortext
+	switch(ftl_start)
```
Qustinnus: Don't use a switch for an if(bool), use if(ftl_start), else

PR #3507, opened 2018-04-02
Add the Scarab back into normal rotation

crossedfall:
This simply renames the Scarab's dmm file to match its respective call so that it can be loaded in without having to manually rename it every-time.

:cl: Crossedfall
fix: Renamed scarab.dmm to **Scrab.dmm**
/:cl:

[why]: # Multiple requests have been seen to add the Scarab back into rotation. This will also make debugging and coding for that map _slightly_ easier since you wont have to manually rename the file every-time you setup a clean clone.


crossedfall:
See #3490 for recent changes which would benefit from this change.

monster860:
Wait.... so the problem this whole time was Linux's case sensitive filenames and Ecolli not realizing that?

crossedfall:
Y..yeah..? Wait, you guys didn't know? I thought this whole time it was done on purpose to remove it from rotation for some reason. Welp. Good news. Its uh, fixed now.

_maps/basemap.dm
```diff
@@ -13,7 +13,7 @@
 #include "map_files\Aetherwhisp\aetherwhisp.dmm"
 #include "map_files\SpaceSHIP\spaceship.dmm"
 #include "map_files\Trailblazer\trailblazer.dmm"
-#include "map_files\Scarab\scarab.dmm"
+#include "map_files\Scarab\Scarab.dmm"
```
crossedfall: Good thinking, makes things appear a bit cleaner- even if it's not directly required.

PR #3526, opened 2018-04-24
Make shells explosive again

### LINKED ISSUE #3524: MAC Cannon shells do not become explosive in their armed state
Currently, regardless of its state, the MAC shells are more-or-less invulnerable. A player can arm the shell, load it into the cannon, ignore all the prompts, eject the shell, and nothing will happen. Same goes for direct hits to armed shells from enemy fire or exploding welder tanks. Tested and confirmed this issue with @Altangy

### Reproduction Steps

1. Arm a shell, doesn't matter which kind
2. Load the shell into the cannon
3. Eject the shell

**_Expected_**: The shell will be ejected into the back-wall and explode
_**Actual**_: The shell will hit the back wall with a thud and can be re-loaded no problem 💣

crossedfall:
This change makes mac shells explode if armed and ejected from the cannon again. I have also increased the explosion radius to make it a little more punishing. The new explosion radius is enough to do some serious damage to munitions and the surrounding rooms, but not enough to be a round ended. The operator has a chance to survive a shell explosion.

:cl: Crossedfall
tweak: Shell explosion size, made it the equivalent of an explosion with a power of 25. _(2, 5, 10)_
fix: Replaced the old "Bump" movement proc with the better "Collision" proc (see [atoms_movable.dm](https://github.com/FTL13/FTL13/blob/master/code/game/atoms_movable.dm) as a reference)
/:cl:

## Old explosion
old explosion
## New explosion
New explosion

[why]: Shells are meant to be explosive when armed. This will resolve #3524

### DIFF
--- a/code/game/objects/structures/cannon_shell.dm
+++ b/code/game/objects/structures/cannon_shell.dm
@@ -29,11 +29,11 @@
 	if(armed) icon_state = "[icon_state]_armed"
 	else icon_state = initial(icon_state)
 
-/obj/structure/shell/Bump(obstacle)
-	if(throwing && armed)
-		explosion(get_turf(src), 1, 2, 6)
-		throwing = 0
-	..()
+/obj/structure/shell/Collide(atom/A)
+	if((A))
+		if(throwing && armed)
+			explosion(get_turf(src), 1, 2, 6)
+			throwing = 0
 
 /obj/item/weapon/twohanded/required/shell_casing
 	name = "cannon shell casing"
--- a/code/game/objects/structures/cannon_shell.dm
+++ b/code/game/objects/structures/cannon_shell.dm
@@ -32,7 +32,7 @@
 /obj/structure/shell/Collide(atom/A)
 	if((A))
 		if(throwing && armed)
-			explosion(get_turf(src), 1, 2, 6)
+			explosion(get_turf(src), 2, 5, 10)
 			throwing = 0
 
 /obj/item/weapon/twohanded/required/shell_casing


code/game/objects/structures/cannon_shell.dm
```diff
@@ -29,11 +29,11 @@
...
-/obj/structure/shell/Bump(obstacle)
-	if(throwing && armed)
-		explosion(get_turf(src), 1, 2, 6)
-		throwing = 0
-	..()
+/obj/structure/shell/Collide(atom/A)
+	if((A))
+		if(throwing && armed)
+			explosion(get_turf(src), 2, 5, 10)
+			throwing = 0
```
Qustinnus: FALSE

Qustinnus:
also, maybe make explosion radius different per shell.

code/game/objects/structures/cannon_shell.dm
crossedfall: T-True..? I need some more input here

crossedfall:
Ah! Good idea. I'll add some new vars for that and apply some defaults to each shell.

LordWalnut:
HE Should be huge.
Huge...

code/game/objects/structures/cannon_shell.dm
Qustinnus: dont use 0 and 1, use FALSE and TRUE

crossedfall: Ah. I kept it that way to coincide with how it's used in atoms_movable where the variable is stored. Although, reverting it to null here would likely be better anyways since that's what it will be reset to naturally.

crossedfall:
Altered the throwing variable to reset to its null state. Tested and confirmed everything still operates as expected. @Qustinnus Let me know if there are any other changes you'd like me to make!

PR #3528, opened 2018-04-24
gives the crew more to do than shoot ships get shot move paper

TMTIME:
instead gives them new missions to sit in a system and shoot ships and move lots of paper!

:cl: TMTIME
add: Trade mission
add: Hold system mission
/:cl:

Adds a small amount of variety to a round with new objectives.

The first is simple. Raise some money. Somewhat easy if they sell tech disks.

The second is wave defence in syndicate space. Eliminate a set number of waves of increasing size.

### DIFF
--- a/code/controllers/subsystem/ships.dm
+++ b/code/controllers/subsystem/ships.dm
@@ -299,7 +299,7 @@ SUBSYSTEM_DEF(ship)
 			O.ships_killed++
 	for(var/datum/objective/ftl/hold_system/O_h in SSstarmap.ship_objectives)
 		if(SSstarmap.current_system == O_h.target_system && O_h.wave_active)
-			O_h.ships_remaining --
+			O_h.ships_remaining--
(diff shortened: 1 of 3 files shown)


code/controllers/subsystem/ships.dm
```diff
@@ -297,6 +297,9 @@ SUBSYSTEM_DEF(ship)
 	for(var/datum/objective/ftl/killships/O in SSstarmap.ship_objectives)
 		if(S.faction == O.faction)
 			O.ships_killed++
+	for(var/datum/objective/ftl/hold_system/O_h in SSstarmap.ship_objectives)
+		if(SSstarmap.current_system == O_h.target_system && O_h.wave_active)
+			O_h.ships_remaining --
```
Qustinnus: why the space between the var and the decrement

Qustinnus: i mean if it works it works just wanna make sure

TMTIME: Old habit I guess? commit up to remove it

Qustinnus:
theres more of them 👀

PR #3532, opened 2018-04-28
Fixes hold objective from always failing + stops floyd calling the polis on me + fixes shuttles being unable to land

TMTIME:
Turns out I don't know how objectives work

:cl: TMTIME
fix: Shuttles can now land on lavaland and ice planets
fix: Hold objective now doesn't fail as soon as the ship leaves the system
/:cl:

### DIFF
--- a/code/game/gamemodes/objective.dm
+++ b/code/game/gamemodes/objective.dm
@@ -1045,6 +1045,8 @@ GLOBAL_LIST_EMPTY(possible_items_special)
 		explanation_text = "Distress signal spoofed to their fleets. Hold [target_system] until operation |REDACTED| is completed. ([current_wave]/[total_waves])."
 
 /datum/objective/ftl/hold_system/proc/manage_waves()
+	if(current_wave >= total_waves)
+		return TRUE
 	if(!holding_system && SSstarmap.current_system == target_system) //They have just arrived, start the waves
 		holding_system = TRUE
 		next_wave_start_time = world.time + rand(500,1000)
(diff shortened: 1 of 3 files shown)


code/game/gamemodes/objective.dm
```diff
@@ -1021,6 +1021,7 @@ GLOBAL_LIST_EMPTY(possible_items_special)
 	var/wave_active = FALSE
 	var/ships_remaining = -1
 	var/next_wave_start_time
+	var/hold_completed = FALSE
```
Qustinnus: dont do this just return 1 if its completed on the check_completion

code/game/gamemodes/objective.dm
```diff
@@ -1060,13 +1061,16 @@ GLOBAL_LIST_EMPTY(possible_items_special)
 				ship_spawned.mission_ai = new /datum/ship_ai/guard
 				ship_spawned.mission_ai:assigned_system = target_system
```
Qustinnus: i will call the police on you

code/game/gamemodes/objective.dm
```diff
@@ -1057,16 +1058,22 @@ GLOBAL_LIST_EMPTY(possible_items_special)
...
-		current_wave = total_waves
+			if(current_wave < total_waves)
+				wave_active = FALSE
+				next_wave_start_time = world.time + rand(100,300)
+				current_wave++
+				update_explanation_text()
+
+/datum/objective/ftl/hold_system/check_completion()
+	manage_waves()
+	if(completed)
```
Qustinnus: Don't use completed. Use if(manage_waves()) and make manage waves return TRUE or FALSE instead of changing the completed var

is this kinda dumb, yes.

but I think it'll be better for code quality to keep it in 2 procs

PR #3534, opened 2018-05-04
Fixes shield penetration, cleans up code and other minor fixes

TMTIME:
now with 100% less X == bitflag!

:cl: TMTIME
tweak: Cleans up my older ship weapon code
balance: Lets AI ships use updated shield damage values
/:cl:

im bad at coding okay

### DIFF
--- a/code/controllers/subsystem/ships.dm
+++ b/code/controllers/subsystem/ships.dm
@@ -119,7 +119,7 @@ SUBSYSTEM_DEF(ship)
 	var/starting_shields = S.shield_strength
 	if(world.time > S.next_recharge && S.recharge_rate)
 		S.next_recharge = world.time + S.recharge_rate
-		S.shield_strength = min(initial(S.shield_strength), S.shield_strength + 1)
+		S.shield_strength = min(initial(S.shield_strength), S.shield_strength + rand(750,1250))
 		if(S.shield_strength >= initial(S.shield_strength))
 			if(S.shield_strength > starting_shields) broadcast_message("<span class=notice>[faction2prefix(S)] ship ([S.name]) has recharged shields to 100% strength.</span>",notice_sound,S)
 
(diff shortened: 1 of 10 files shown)


code/datums/ship.dm
```diff
@@ -466,6 +468,14 @@ GLOBAL_VAR(next_ship_id)
 
 	fire_rate = 500
 
+/datum/ship_component/weapon/unknown_ship_weapon/attack_effect(var/turf/T) //10 shots, 7 spread
+	var/turf/target_sub
+	new /obj/effect/temp_visual/ship_target(T, attack_data) //Initial hit
```
Qustinnus: There's a pause of 5 seconds before this actually does an effect

code/controllers/subsystem/ships.dm
```diff
@@ -236,8 +227,8 @@ SUBSYSTEM_DEF(ship)
 		spawn(10)
 			if(istype(S)) // fix for runtime (ship might have ceased to exist during the spawn)
 				broadcast_message("<span class=notice>Shot hit! ([S.name])</span>",success_sound,S)
-	if(S.shield_strength >= 1 && !attack_data & SHIELD_PENETRATE)
-		S.shield_strength = max(S.shield_strength - attack_data.hull_damage, 0)
+	if(S.shield_strength >= 1 && !(attack_data.unique_effect & SHIELD_PENETRATE))
```
Qustinnus: Make this like 100 or 250

code/controllers/subsystem/ships.dm
```diff
@@ -119,7 +119,7 @@ SUBSYSTEM_DEF(ship)
 	var/starting_shields = S.shield_strength
 	if(world.time > S.next_recharge && S.recharge_rate)
 		S.next_recharge = world.time + S.recharge_rate
-		S.shield_strength = min(initial(S.shield_strength), S.shield_strength + 1)
+		S.shield_strength = min(initial(S.shield_strength), S.shield_strength + rand(750,1250))
```
Qustinnus: 750 - 1250 is a stupid amount and would be OP on some ships, make this a ship var or use components in conjunction with this instead of a magic number

TMTIME: 750-1250 is the same as what it was, just with a little random variance. It used to just regen 1 point every repair phase.

Also not really OP since most ships are killed before even getting a repair tick due to OP cells.

Qustinnus: Yeah but were removing OP cells, we should code for the future

PR #3538, opened 2018-05-05
Ineese Map Changes 2

IndusRobot:
[Changelogs]: # (Please make a changelog if you're adding, removing or changing content that'll affect players. This includes, but is not limited to, new features, sprites, sounds; balance changes; map edits and important fixes)

:cl: TMTIME
add: Added round shields. Extra space area was used to fit the shields. 
fix: Removed the last active roundstart turf. Probably. 
add: Added an emergency shuttle manipulator to the console for those speedy escapes. Don't worry, a pod has already been provided for the bridge. 
add: Added a slime processor and slime management console in maintenance beside science. 
add: More emergency toolboxes. 
fix: Moved the detective's locker out of the armory to somewhere accessible but still secure. 
add: Added XO desk shutters
add: Added a cargo shuttle. Backup console is located on the bridge. 
add: Added holopads. 
/:cl:

[why]: # (Please add a short description [on the next line] of why you think these changes would benefit the game. If you can't justify it in words, it might not be worth adding:)

PR #3539, opened 2018-05-05
Meteors now work with shields

TMTIME:
help im coding too much

:cl: TMTIME
add: Sounds effect for shields dropping is now in
fix: Shields and meteors now interact properly
/:cl:

also works with meaty ores and tunguska

Only merge with a PR like #3538, due to no ships currently having shields. no point bringing down the server to add nothing


Qustinnus:
dont make shields start empty, they're not empty at start for a reason.

PR #3557, opened 2018-05-20
Doubles the gas penalties and increases explosion power

ncaaseawolf:
Like the above says. I feel like the "boom" from a SM exploding is not big enough and incompetent engineers should face harsher punishment.

### DIFF
--- a/code/modules/power/supermatter/supermatter.dm
+++ b/code/modules/power/supermatter/supermatter.dm
@@ -2,10 +2,10 @@
 //Please do not bother them with bugs from this port, however, as it has been modified quite a bit.
 //Modifications include removing the world-ending full supermatter variation, and leaving only the shard.
 
-#define PLASMA_HEAT_PENALTY 15     // Higher == Bigger heat and waste penalty from having the crystal surrounded by this gas. Negative numbers reduce penalty.
-#define OXYGEN_HEAT_PENALTY 1
-#define CO2_HEAT_PENALTY 0.1
-#define NITROGEN_HEAT_MODIFIER -1.5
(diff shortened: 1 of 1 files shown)


code/modules/power/supermatter/supermatter.dm
```diff
@@ -2,10 +2,10 @@
...
 //Modifications include removing the world-ending full supermatter variation, and leaving only the shard.
 
-#define PLASMA_HEAT_PENALTY 15     // Higher == Bigger heat and waste penalty from having the crystal surrounded by this gas. Negative numbers reduce penalty.
-#define OXYGEN_HEAT_PENALTY 1
-#define CO2_HEAT_PENALTY 0.1
-#define NITROGEN_HEAT_MODIFIER -1.5
+#define PLASMA_HEAT_PENALTY 30     // Higher == Bigger heat and waste penalty from having the crystal surrounded by this gas. Negative numbers reduce penalty.
+#define OXYGEN_HEAT_PENALTY 2
+#define CO2_HEAT_PENALTY 0.2
+#define NITROGEN_HEAT_MODIFIER -3
```
crossedfall: #### Plasma: 
I expect this to impact lower-end engine setups substantially, making current "learning" builds nonviable. N20's values were left unchanged, I'd be interested in knowing if the unchanged gas can still cope with your proposed changes to plasma specifically.

#### N2
Have you looked into power outputs from builds using N2? I'd also imagine that the healing factor of N2 is pretty crazy now.

#### O2: 
Any reason to actually double the o2comp value when determining the dynamic heat modifier? I expect that in conjunction with Plasma the engine will be a lot more volatile. What gas mixes have you tested with these changes?

Qustinnus:
the SM explosion just punishes the crew and I already think its a shit feature, this PR would only make that worse.

PR #3559, opened 2018-05-28
Properly does quieter sounds

TMTIME:
:cl: TMTIME
tweak: Lowered the volume for Dolos but please don't go there okay
/:cl:

shhh


Qustinnus:
Remove line 132 in starmap.dm

PR #3579, opened 2018-06-11
Changes random name generation to use the ship name generation

Vivalas:
As the title says

:cl: Vivalas
tweak: Station name is now picked using the same method NPC ship names are picked.
/:cl:

[why]: Because NSV Buttmasher is better than Hooligan Waypoint Seven

### DIFF
--- a/code/__HELPERS/names.dm
+++ b/code/__HELPERS/names.dm
@@ -60,7 +60,7 @@ GLOBAL_VAR(command_name)
 		if(config && config.station_name)
 			newname = config.station_name
 		else
-			newname = new_station_name()
+			newname = "NSV [generate_ship_name()]"
 
 		set_station_name(newname)
 


code/__HELPERS/names.dm
```diff
@@ -60,7 +60,7 @@ GLOBAL_VAR(command_name)
 		if(config && config.station_name)
 			newname = config.station_name
 		else
-			newname = new_station_name()
+			newname = "NSV [generate_ship_name()]"
```
crossedfall: Should we have a secondary variable to set the ship class? It would be pretty nice if we could do it based on the ship-map selected for the round.

If not, given that most of our missions are combat related it doesn't make much sense to send a Nanotrasen Science Vessel.  Maybe we should use a different standard classification NSC [Nanotrasen Space Cruiser], NSS [Nanotrasen Space Ship], NAC [Nanotrasen Assault Cruiser], etc.

Vivalas: Well I generally use Bay12 designations, so the S is indeed science, but in this case it could probably be abstracted to mean NanotraSen Vessel, which is what I really intend anyways.

crossedfall: That.. Doesn't make sense, but alright. s:

Qustinnus: @Vivalas then itd be NTV you dumbass

PR #3583, opened 2018-06-12
Plasma glass window construction fix.

Metallumere:
:cl: Metallumere
fix: Fixed clicking on grills with plasma sheets not constructing windows, it now makes the proper window you would expect.
fix: Fixed clicking reinforced plasma sheets on themselves not doing anything, they now open the construction window you would expect.
/:cl:


Qustinnus:
seems ok

PR #3585, opened 2018-06-21
Ports TG's sleeping code, minus trait checks.

Metallumere:
Not sure if this fix is all-encompassing. I don't know if sleep properly makes you less drunk and such, but otherwise it works for stopping falling over and getting back up immediately upon being proc'ed.
It's a straight copypaste of TG's current code, but without the trait checks.

:cl: Metallumere
fix: Previously the 'sleep' command and n2o put you to sleep, but you immediately woke up. That's fixed.
/:cl:

[why]: 
So anesthetic actually works, and having an n2o breach is something you should actually deal with sooner rather than never-because-it's-not-a-problem.

### DIFF
--- a/code/modules/mob/status_procs.dm
+++ b/code/modules/mob/status_procs.dm
@@ -71,28 +71,24 @@
 
 /////////////////////////////////// SLEEPING ////////////////////////////////////
 
-/mob/living/proc/IsSleeping() //If we're asleep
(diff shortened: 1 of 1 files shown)


Metallumere:
I've tested this a bit more extensively by someone else's suggestion of things to test. It would seem everything functions as it should. I am faaaaaaairly certain nothing is broken.

PR #3590, opened 2018-07-06
The Butterfly, a new small (but not too small) ship!

NIXC:
Issue tracking sheet: 
https://docs.google.com/document/d/1k_JVqzL70hTaMG6yIUkSJt9rf8l7Zb9zLoesO1bcgyw/edit?usp=sharing

:cl:  Butterfly
add: The Butterfly, a new small vessel utilizing solars.

/:cl:
ya


NIXC:
Apparently access txt is not supposed to have "as text" after when you varedit it. that seems to be the problem.

PR #3595, opened 2018-07-11
Fixes parallax shuttle animation on the ship

monster860:
:cl: monster860
fix: Finally, the ship actually feels like it's moving! The parallax now does the whole scrolling thing.
/:cl:


Qustinnus:
the parallax change hits too soon and isnt in sync with audio

PR #3602, opened 2018-07-21
Ship name fix

MadmanMartian:
Ships names now have a space between the prefix and the suffix, no more ClownBlaster and TitanSmasher, instead it's Titan Smasher.

:cl: 
fix: Fixes random ship name generation so there is a space between the prefix and the suffix
/:cl:

[why]: # I was playing and saw the ship names coming up as TitanSmasher, FaceBlaster, and the like.


MadmanMartian:
How on earth did I manage to conflict with that? That's what I was changing!

Qustinnus:
fixed it for ya

MadmanMartian:
Thanks. I must have an old fork I was unaware of

LordWalnut:
image
 
I don't like the look of that.

MadmanMartian:
Oh, it's double spaced for some reason. Odd.

NIXC:
Its because the string has two spaces in it....

Also, the suffix should probably be capitalized in this case.

MadmanMartian:
I was talking about the string having two spaces in it. Could have sworn I only put one.

I'll see what I can do about the suffix

PR #3609, opened 2018-08-11
Station/ship destroyed, mission failed.

MadmanMartian:
Should the ship receive critical damage, the round ends.

:cl: 
add: Should the ship receive critical damage. The round ends.
/:cl:

[why]: # Right now, should the ship get turboclucked into the 8th ring of hell with nobody alive to call it off, it resolves to admin intervention to end the round. This is somewhat unfortunate should an admin not be present to press the explode-everything button.

### DIFF
--- a/code/game/gamemodes/game_mode.dm
+++ b/code/game/gamemodes/game_mode.dm
@@ -186,6 +186,14 @@
 		return TRUE
 	if(station_was_nuked)
 		return TRUE
+
+	if(config.check_station_integrity)
+		var/static/datum/station_state/end_state = new /datum/station_state() //Static so it doesn't have to keep being remade.
+		end_state.count()
+		var/station_integrity = min(PERCENT(GLOB.start_state.score(end_state)), 100)
+		if(station_integrity <= 20) //Below 20% integrity
+			message_admins("The station/ship has suffered critical damage, with the config set so that it ends the round with critical damage. Doing so..")
+			return TRUE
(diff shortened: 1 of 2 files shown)


code/game/gamemodes/game_mode.dm
```diff
@@ -186,6 +186,14 @@
 		return TRUE
 	if(station_was_nuked)
 		return TRUE
+
+	if(config.check_station_integrity)
+		var/static/datum/station_state/end_state = new /datum/station_state() //Static so it doesn't have to keep being remade.
+		end_state.count()
+		var/station_integrity = min(PERCENT(GLOB.start_state.score(end_state)), 100)
+		if(station_integrity <= 20) //Below 20% integrity
+			message_admins("The station/ship has suffered critical damage, with the config set so that it ends the round with critical damage. Doing so..")
```
crossedfall: This is a little awkwardly phrased. I don't think you need to expressly tell the admins that the round is going to end due to a config setting. I think it's obvious enough from the fact that they're being messaged about it. Also, it's always going to be a ship, so you might as well specify as such.

Maybe: "The ship has sustained critical damage! Ending the round."

MadmanMartian: That does sound better

Qustinnus:
I'm pretty sure integrity checks are super expensive

MadmanMartian:
oh damn it really is. It loops through every turf on z level 1

PR #3610, opened 2018-08-11
Hostile ship fire does not logspam admins anymore

MadmanMartian:
[why]: # After watching a round where the staff jumped to the syndicate capital system, I noted that the admins were having trouble getting the story because of the spam of fire from the syndicate capital ships.


crossedfall:
Duplicate of #3562

PR #3615, opened 2018-09-08
Makes gas mouse transparent

ninjanomnom:
:cl: ninjanomnom
tweak: Gas overlays no longer block clicks when a stray pixel goes in front of your mouse
/:cl:

An easy to port tg thing you guys apparently don't have. This does require 512.

Port of https://github.com/tgstation/tgstation/pull/32569 and other changes to the same feature


ninjanomnom:
If you want travis to work for this you're going to need to update it to 512

PR #3621, opened 2018-09-21
Immortalize some important people

crossedfall:
This will ensure that some important people are always remembered

:cl: Crossedfall
add: New possible random ship names
/:cl:

[why]: # Because remembering is important.


crossedfall:
Gotta remember who made the first version of FTL13 even remotely possible.

TMTIME:
yeah i died

PR #3627, opened 2018-10-20
New smol map, Yacht! (Needs tgm conversion do not merge)

Altangy:
[Changelogs]:
:cl: Altangy
add: Yacht! The newest smallest non-joke ship in the fleet!
add: Yacht is in the config as well. Anyone with that config excluded will need to add it.
experiment: Yacht evade is at 90%. In testing I've noticed that it still gets hit often.
/:cl:

[why]: # (Please add a short description [on the next line] of why you think these changes would benefit the game. If you can't justify it in words, it might not be worth adding:) 

So yea, it's ready other than a tgm convert which my computer won't do and ninja talk about shortly.


Altangy:
So I'm trying to find why travis failed and all I can see is it stalled.... no idea...

crossedfall:
Restarted the build and it looks to have compiled. 👍

_maps/Yacht.json
```diff
@@ -0,0 +1,27 @@
+{
+	"map_name": "Yacht",
+	"map_path": "map_files/Yacht",
+	"map_file": "Yacht.dmm",
+	"minetype": "lavaland",
+	"transition_config": "default",
+
+	"player_ship_hit_chance": "0.10",
```
TMTIME: 10% base chance to hit the players is beyond op

Altangy: If it gets hit once it's basically dead, I'd be willing to drop it to 50 or 40.  Also, I have multiple late joins and don't see where I could fit more. I'll look into fixing the terminal on the shield it should be connected.

Altangy: Also I should point out that the fake lattice blocks don't come with the ship when it jumps. I'll remove them anyway because they are dumb

TMTIME: 60 is fine since the FOBshuttle already uses that.

I'm not on about adding more chairs, just replacing the area that room is in with /area/shuttle/ftl/hallway/secondary/entry
tldr first the game checks latejoin landmarks and if they have a chair in them
if they don't, it just throws them anywhere in that area
if that area doesn't exist it just leaves them in the title room

_maps/Yacht.dm
```diff
@@ -0,0 +1 @@
+#define FORCE_MAP "_maps/Butterfly.json"
```
NIXC:
```suggestion
#define FORCE_MAP "_maps/Yacht.json"
```
I'll just leave this here.

config/maps.txt
```diff
@@ -36,4 +36,6 @@ map Feliciana
 endmap
 
 map Butterfly
+endmap
+map Yacht
```
crossedfall:
```suggestion

map Yacht
```
add a break to maintain formatting and ease of readability.

_maps/Yacht.dm
Altangy: Huh... when I made the commit it was like that... thanks for catching it though

_maps/Yacht.json
Altangy: Alright, I fixed everything you asked for, it is not reflected perfectly in that picture

PR #3628, opened 2018-10-23
Updates to the latest 512 version

crossedfall:
:cl: Crossedfall
tweak: Updated travis to use the latest 512 version to reduce compile times.
/:cl:

[why]: Even if we're only submitting major bug fixes and some new map stuff to this repo, updating Travis can help make the testing process go faster


TMTIME:
nobody saw that

PR #3630, opened 2018-11-15
added the ability for the shield generator to be upgraded

ondrej008:
:cl: Ondrej008
add: The FTL Shield Generator can now be upgraded, a capacitor increases the energy charge rate, max energy stored, max plasma stored, and a manipulator increases the plasma charge rate.
/:cl:

This isn't really a great solution, but it should atleast improve some things.
Generator can be deconstructed now, not sure if that's a good idea, but it's there.
Mess with the values of the upgrades, if you think it'd make them better.

This is my first PR on here, so I'm sorry if I did something wrong.

### DIFF
--- a/code/modules/ftl/machinery/ftl_shieldgen.dm
+++ b/code/modules/ftl/machinery/ftl_shieldgen.dm
@@ -1,5 +1,3 @@
-// ondrej008 code, experimental!!!
-
 /obj/machinery/ftl_shieldgen
 	name = "Shield Generator"
 	desc = "A big ass thing to make sure the clowns don't turn your ship into banana cream pies."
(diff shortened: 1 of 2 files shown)


code/modules/ftl/machinery/ftl_shieldgen.dm
```diff
@@ -93,6 +125,14 @@
 	update_icon()
 	update_active()
 
+/obj/machinery/ftl_shieldgen/attackby(obj/item/weapon/W, mob/user, params)
```
Qustinnus: Im not really sure how this part works, can you elaborate? besides that. this PR should be fine.

ondrej008: If you mean the attackby(...), then I added it, so that the parts could be exchanged, and so that the shieldgen could be deconstructed. And now I just noticed the deconstruction screwdriver thingy with ore_redemption, huh.

PR #3632, opened 2018-11-23
FTL drive can now be upgraded

ondrej008:
🆑 Ondrej008
add: The FTL Drive can now be upgraded, a capacitor increases the range that you can jump to, manipulators increase the FTL speed.
/🆑

The FTL Drive now can be deconstructed and upgraded. The manipulators can increase the speed by 4 times. Capacitors can increase the jump range to 38.
Basically, it takes 35s + ftl warmup to jump to a different star system now, compared to about 2mins + ftl warmup. I thought about making the time be dependant on distance to the system, but that can be done later. The hardest part of this was making the tgui thing look nice.

You can change the values to be more reasonable, if you want to.


code/modules/ftl/machinery/ftl_drive.dm
```diff
@@ -50,6 +81,8 @@
 	if(!istype(get_area(src), /area/shuttle/ftl) || (SSstarmap.ftl_drive && isturf(SSstarmap.ftl_drive.loc)))
 		stat |= BROKEN
 		return
+	var/obj/item/weapon/circuitboard/machine/ftl_drive/B = new
```
Qustinnus: What will this accomplish? The circuitboard will be removed once Initialize() is over

ondrej008: it applies the default parts to the machine, otherwise I think the machine would be there, but it wouldn't have the stock parts in it (the tier 1 research parts like manips, caps)

PR #3640, opened 2018-12-10
Adds some nice debug messages to the FTL drive, maybe I can finally figure out why it doesn't work on the server.

ondrej008:
:cl: ondrej008
add: Debug messages to the FTL drive.
/:cl:

So that I can fix the damn FTL drive, maybe. Also am I supposed to make a new branch on my fork of this for each new change or how should it work?


code/controllers/subsystem/starmap.dm
```diff
@@ -177,7 +178,7 @@ SUBSYSTEM_DEF(starmap)
 	if(timeleft > 0)
 		return "[add_zero(num2text((timeleft / 60) % 60),2)]:[add_zero(num2text(timeleft % 60), 2)]"
 	else
-		return "00:00"
+		return "00:[timeleft]"
```
Qustinnus: this'll be 00:0

code/controllers/subsystem/starmap.dm
```diff
@@ -199,6 +200,14 @@ SUBSYSTEM_DEF(starmap)
 		return 1
 	if(ftl_is_spooling)
 		return 1
+	ftl_message("<span class=notice>FTL Drive attempting jump to [target.name], a [target.alignment] system. \
+		[ftl_drive.jump_speed == initial(ftl_drive.jump_speed) ? "" : " FTL drive upgrades report a speed of [1/ftl_drive.jump_speed] times faster travel."] \
+		[ftl_drive.jump_max == initial(ftl_drive.jump_max) ? "" : " Jump range increased by [ftl_drive.jump_max - initial(ftl_drive.jump_max)] light years."] \
+		</span>")
+	spawn(30)
+		ftl_message("<span class=notice>Calculations indicate that we should arrive at our destination in roughly [((1850 * ftl_drive.jump_speed) + 600) / 10] seconds, and we should be in FTL for [(1850 * ftl_drive.jump_speed) / 10] seconds.</span>")
+	spawn((1850 * ftl_drive.jump_speed) + 600 - 50) // the - 50 is there because otherwise it would kind of say that once you jump out
```
Qustinnus: use timers instead of spawn

ondrej008: there's literally a spawn everywhere in that proc, 2 more can't hurt, right?
i'll see what I can do about timers

Qustinnus: if theres poop on your desk pooping on your desk even more doesnt become ok

ondrej008: ok I put the spawn((1850... in a timer, is the spawn(30) okay to leave?
why should I put it in a timer anyway? what's wrong with spawn?

Qustinnus: ok imagine you order pizza and it'll be there in 30 minutes so you tell yourself "wait 30 minutes".

only a retard (byond) would stand at the door for 30 minutes like wow i hope my pizza is almost here

a smart person is like yes i will come back to the door in 30 minutes to run the open door function (a timer)

ondrej008: oh okay, so it just pauses the current thread, whereas a timer just gives the thread to other things

Also is it fine to have the spawn(30) or should I turn that into a timer too?

ondrej008: @Qustinnus is it good then?

Qustinnus: Yeah sure btu in the future try to use timers for everything instead of sleep or spawn :P

ondrej008: @Qustinnus ok but won't I have to make a proc for each timer? Like the proc in the SS starmap, which uses a crap ton of spawns, would all those spawns need to be put into separate procs?

Qustinnus: we dont for the starmap because its not a super common thing and theyre short, but for other things, you would have to make new procs.

PR #3644, opened 2018-12-15
Reduce Event Player Requirements

Karmic-Skink:
Reducing player number requirements for some already existing events to occur, given current average player numbers.

Disease_outbreak:
Min Players 10 ->5 //Likely be a virologist at that point anyway
Portal_storm:
Min Players 15 ->10 Max Occurrences = 1
Processor Overload:
Min Players 20 ->5
Space Vine:
Min Players 10 ->5  Max Occurrences 3 -> 1
Spider Infestation:
Min Players 15 ->10


Altangy:
The lowest any of these events should trigger at is 7-9 players. I've administrated 5 person rounds and I can tell you that most of the time,most of the crew have no idea what they are doing and only one person actually knows the difference between space and the floor.

crossedfall:
For the portal storms it might be good to add a check to ensure that it only procs in syndicate space. Doesn't make too much sense for them to proc while the ship is in friendly space, if we're honest. As for the actual reduced counts, I'll defer to the admins who have been overseeing rounds for longer than I have.

Funce:
I only am taking issue with reducing the disease outbreak number. A virologist isnt exactly what you'd call standard in a round of 5-9 people.

If you're relying on the fact that medical may have access, that's a little bit power gamey.

Other numbers are fine.

Vivalas:
I second the opinion that the lowest should be 10 players

Karmic-Skink:
Honestly, I just wanted to start a conversation about these. Sometimes low-pop without admins gets really dull with only a few events triggering.

Are there any that you'd be happy lowering? I purposely left out things like immovable rod, given it will delete things like the FTL drive.

I've got a fix for the syndicate version of the portal storm, but it currently applies to the narsie version too since they are in the same file. I'll keep trying to work that one out. But yeah, I'm not an experienced coder.

PR #3652, opened 2018-12-24
Improves the antimatter!

ondrej008:
:cl: Ondrej008
tweak: Antimatter stability auto-updates, now the antimatter won't show 100% stability and still explode!
tweak: If you place the antimatter shielding incorrectly, it won't melt into thin air, rather it goes back into the packaging!
tweak: the antimatter containment jar (fuel for antimatter) now shows how many units there are left!
tweak: the antimatter controller menu looks nicer and auto refreshes!
tweak: the antimatter doesn't make a sound each time it makes power, it also has different sounds!
/:cl:

The jar thing is QoL, the only way to know how much was in one was to put it inside an antimatter control, now you can examine that thingy.

The stability is really useful, before it could say 100% stability on both things and still explode, there was no way to tell if the AM is taking damage or exploding without clicking the damn (update) button, and the Stability: was misleading.

The shielding is also QoL and makes more sense than having unpowered, deactivated machinery just melt.

Dropping those PRs like my mixtape.


ondrej008:
If someone knows how to make the antimatter control unit menu auto-refresh, please do tell!

code/modules/power/antimatter/control.dm
```diff
@@ -85,16 +88,26 @@
 	if(fuel > (2*core_power))//More fuel has been put in than the current cores can deal with
 		if(prob(50))
 			core_damage = 1//Small chance of damage
+			playsound(src.loc, 'sound/effects/bang.ogg', 25, 1)
 		if((fuel-core_power) > 5)
 			core_damage = 5//Now its really starting to overload the cores
+			playsound(src.loc, 'sound/effects/bang.ogg', 50, 1, 5)
 		if((fuel-core_power) > 10)
 			core_damage = 20//Welp now you did it, they wont stand much of this
+			playsound(src.loc, 'sound/effects/bang.ogg', 200, 1, 10) // 5 of these and it gibs in 7 range
```
Qustinnus: does 200 even do anything

ondrej008: good question! i'll go try it right now!

code/modules/power/antimatter/shielding.dm
```diff
@@ -27,20 +27,28 @@
...
 	. = ..()
+	destroy_message = "<span class='danger'>The [name] melts!</span>"
 	addtimer(CALLBACK(src, .proc/controllerscan), 10)
 
 
 /obj/machinery/am_shielding/proc/controllerscan(priorscan = 0)
 	//Make sure we are the only one here
 	if(!istype(src.loc, /turf))
+		new/obj/item/device/am_shielding_container(src.loc)
+		visible_message("This should never happen! src.loc wasn't a type of /turf in shielding.dm")
```
Qustinnus: theres a different proc for this IIRC

code/modules/power/antimatter/shielding.dm
```diff
@@ -27,20 +27,28 @@
...
+		destroy_message = "<span class='notice'>The [name] folds back into it's package.</span>"
 		qdel(src)
 		return
 	for(var/obj/machinery/am_shielding/AMS in loc.contents)
 		if(AMS == src)
 			continue
+		new/obj/item/device/am_shielding_container(src.loc)
+		visible_message("<span class='warning'>Cannot have more than one shielding in the same place.</span>")
```
Qustinnus: ditto

code/modules/power/antimatter/shielding.dm
ondrej008: you mean the new/obj...? not sure about that, I just used that because that's all I know

Qustinnus: no. visible message. this is an error message so it shouldnt appear to the player

PR #3653, opened 2018-12-25
Adds compact_units(amount, unit) and uses it in multitool.

ondrej008:
:cl: Ondrej008
tweak: multitool now displays in W, kW, MW depending on the power
/:cl:

Looks nicer.


code/modules/power/powernet.dm
```diff
@@ -96,4 +96,29 @@
...
+
+example:
+compact_units(50000,    "W") -> 50 kW
+compact_units(50000000, "W") -> 50 MW
+
+*/
+/proc/compact_units(amount, unit)
+	if(amount < 1000)
+		return "[amount] [unit]"
+	if(amount > 1000*1000*1000*1000) // more readable than 1000000000
```
Qustinnus: there might be defines for this

ondrej008: I changed it to use the power thingy **

code/modules/power/powernet.dm
```diff
@@ -96,4 +96,29 @@
...
+
+example:
+compact_units(50000,    "W") -> 50 kW
+compact_units(50000000, "W") -> 50 MW
+
+*/
+/proc/compact_units(amount, unit)
+	if(amount < 1000)
+		return "[amount] [unit]"
+	if(amount > 1000**4) // more readable than 1000000000
```
Qustinnus: yeah but im still pretty sure theres defines

ondrej008: any idea where I should look? I tried misc.dm and math.dm in __DEFINES but nothing so far.

Qustinnus:
eh w.e

PR #3650, opened 2018-12-23
Character slots and emoji parsing updated

MASMC:
Emoji parsing honestly probably didn't do anything.

[Changelogs]: # (Please make a changelog if you're adding, removing or changing content that'll affect players. This includes, but is not limited to, new features, sprites, sounds; balance changes; map edits and important fixes)
[]: # (See here for how to easily make a changelog: https://github.com/tgstation/tgstation/wiki/Changelogs. An example changelog has been provided below. Please edit or remove)

:cl: MASMC
:tweak: Tweaked emoji slots, to be a few higher
:tweak: Added a single emoji line clear to the emoji parser 
/:cl:

[why]: # (Please add a short description [on the next line] of why you think these changes would benefit the game. If you can't justify it in words, it might not be worth adding:) 
Three character slots is a bit limiting. Also added more to the member, of course by the same amount.
The emoji line is just because it fixed it on my own local repo, probably won't on the server.

PR #3655, opened 2018-12-29
FIXES THOSE DAMN EMOJIS! AHHAHAHA, also fixes the bicon_cache thingy

ondrej008:
:cl: Ondrej008
fix: more than one emoji is now possible!
/:cl:

The /tg/ people tried to pass a raw Icon to something that only takes text. They then also failed at copy & paste, because their "cache" wasn't actually a cache and was a new list every time that function got called.


crossedfall:
Closing as this will need to be merged into #3650 s branch

Issue #3658, opened 2019-01-20
Duffelbags are broken, and have been broken, FOR TWO YEARS!!


SgtxRootbeer:
so
in lines 331 and 332 in 
code/game/objects/items/weapons/storage/backpack.dm 
refers to duffelbags icon and item states as "duffle"
this is ok since storage.dmi also has the icon named "duffle"
but fails when you go to:
icons/mob/back.dmi
the issue being that someone didn't check the spelling and the icons are named "duffel"

the real ironic part is that the DMI was changed, in a way that broke it, because our bot mirrored a pr from /tg/.
the /tg/ pr was made to fix an issue
an issue that I MYSELF REPORTED
![image](https://user-images.githubusercontent.com/22112255/51446246-464b0480-1cc4-11e9-9356-7246ed2b023f.png)

crossedfall:
Fixed by #3661

PR #3661, opened 2019-01-22
Duffel bag reference changes

AdrienTheMan:
renamed some of the duffelbag stuff to fix this #3658 

:cl: 
fix: changed duffle to duffel for normal, captain, sec, med, eng, drone duffel bags in their item and icon states
/:cl:


AdrienTheMan:
there we go, reverted the back.dmi stuff and changed the icon_state and item_state of duffles to duffel

AdrienTheMan:
this gets me more comfy with github and dreammaker in the process so no worries, but yes it should be good now

PR #3665, opened 2019-02-04
NPC ship based Chemical Weapons

Karmic-Skink:
:cl: Karmic Skink
add: A ship based shield-piercing chemical weapon in various toxin flavours for use by NPCs against the NSV Crew
add: 3 versions of the weapon mount (slow/normal/fast)
tweak: added this weapon to the s_weapon pool
/:cl:

[why]:
I'd like to spice up ship-to-ship combat a little with a different sort of hazard, one that can be mitigated if the Crew is adequately prepared (or paying attention), but not completely ignored. The aim is to create a weapon that will area deny for a period of 10-15s and cause those exposed to leave their stations to seek medical attention, hence further disrupting ship effectiveness and also give something for medical to do. 

The toxins used currently are all lethal or "lethal" over a moderate period of time, but shouldn't immediately kill any Crew members.

Currently the weapon will only spawn on anything using s_weapon, ie syndi bulkers and NT battle cruisers. This weapon has not been directly assigned to any ships, but remains accessible via the admin FTL console.


code/datums/ship.dm
```diff
@@ -433,6 +433,29 @@ GLOBAL_VAR(next_ship_id)
...
 
+			//Chemical Bombs
+/datum/ship_component/weapon/chembomb
+	name = "chembomber"
+	cname = "chembomber"
+	fire_rate = 300
+
+	attack_data = /datum/ship_attack/chem_bomb
+
+/datum/ship_component/weapon/chem_stunbomb
```
Qustinnus: why the weird typepath

code/datums/ship.dm
```diff
@@ -433,6 +433,29 @@ GLOBAL_VAR(next_ship_id)
 	attack_data = /datum/ship_attack/stun_bomb
 
 
+			//Chemical Bombs
+/datum/ship_component/weapon/chembomb
```
Qustinnus: this is the only one being used by a ship btw, the rest are unused

